PR #3, opened 2020-09-08
Add Fake LoadTest provider

aleksandrzhiliaev:
**Description:**
In order to abstract Kangal runner from one JMeter provider, we've added a second "Fake" provider which runs only one master job and finished after 10 seconds. This gives us opportunity to run Kangal Controller Integration tests faster and plan better architecture with other providers which we want to add besides JMeter.

### DIFF
--- a/pkg/backends/backend.go
+++ b/pkg/backends/backend.go
@@ -30,7 +30,7 @@ type LoadTestType interface {
 
 // NewLoadTest returns a new LoadTestType
 func NewLoadTest(loadTest *loadTestV1.LoadTest, kubeClientSet kubernetes.Interface, kangalClientSet clientSetV.Interface, logger *zap.Logger, namespacesLister coreListersV1.NamespaceLister, reportConfig report.Config, podAnnotations, namespaceAnnotations map[string]string) (LoadTestType, error) {
-	switch ltType := loadTest.Spec.Type; ltType {
+	switch loadTest.Spec.Type {
 	case loadTestV1.LoadTestTypeJMeter:
 		return jmeter.New(kubeClientSet, kangalClientSet, loadTest, logger, namespacesLister, reportConfig, podAnnotations, namespaceAnnotations), nil
 	case loadTestV1.LoadTestTypeFake:
--- a/pkg/backends/backend.go
+++ b/pkg/backends/backend.go
@@ -2,6 +2,7 @@ package backends
 
 import (
 	"context"
+	"fmt"
 
 	"go.uber.org/zap"
 	"k8s.io/client-go/kubernetes"
@@ -28,11 +29,12 @@ type LoadTestType interface {
 }
 
 // NewLoadTest returns a new LoadTestType
-func NewLoadTest(loadTest *loadTestV1.LoadTest, kubeClientSet kubernetes.Interface, kangalClientSet clientSetV.Interface, logger *zap.Logger, namespacesLister coreListersV1.NamespaceLister, reportConfig report.Config, podAnnotations, namespaceAnnotations map[string]string) LoadTestType {
+func NewLoadTest(loadTest *loadTestV1.LoadTest, kubeClientSet kubernetes.Interface, kangalClientSet clientSetV.Interface, logger *zap.Logger, namespacesLister coreListersV1.NamespaceLister, reportConfig report.Config, podAnnotations, namespaceAnnotations map[string]string) (LoadTestType, error) {
 	switch ltType := loadTest.Spec.Type; ltType {
 	case loadTestV1.LoadTestTypeJMeter:
-		return jmeter.New(kubeClientSet, kangalClientSet, loadTest, logger, namespacesLister, reportConfig, podAnnotations, namespaceAnnotations)
-	default:
-		return fake.New(kubeClientSet, loadTest, logger)
+		return jmeter.New(kubeClientSet, kangalClientSet, loadTest, logger, namespacesLister, reportConfig, podAnnotations, namespaceAnnotations), nil
+	case loadTestV1.LoadTestTypeFake:
+		return fake.New(kubeClientSet, loadTest, logger), nil
 	}
+	return nil, fmt.Errorf("load test provider not found: %s", loadTest.Spec.Type)
 }
--- a/pkg/backends/backend.go
+++ b/pkg/backends/backend.go
@@ -7,6 +7,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	coreListersV1 "k8s.io/client-go/listers/core/v1"
 
+	"github.com/hellofresh/kangal/pkg/backends/fake"
 	"github.com/hellofresh/kangal/pkg/backends/jmeter"
 	loadTestV1 "github.com/hellofresh/kangal/pkg/kubernetes/apis/loadtest/v1"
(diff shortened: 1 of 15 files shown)


s-radyuk:
+1

pkg/backends/backend.go
```diff
@@ -27,6 +29,12 @@ type LoadTestType interface {
 }
 
 // NewLoadTest returns a new LoadTestType
-func NewLoadTest(loadTest *loadTestV1.LoadTest, kubeClientSet kubernetes.Interface, kangalClientSet clientSetV.Interface, logger *zap.Logger, namespacesLister coreListersV1.NamespaceLister, reportConfig report.Config, podAnnotations, namespaceAnnotations map[string]string) LoadTestType {
-	return jmeter.New(kubeClientSet, kangalClientSet, loadTest, logger, namespacesLister, reportConfig, podAnnotations, namespaceAnnotations)
+func NewLoadTest(loadTest *loadTestV1.LoadTest, kubeClientSet kubernetes.Interface, kangalClientSet clientSetV.Interface, logger *zap.Logger, namespacesLister coreListersV1.NamespaceLister, reportConfig report.Config, podAnnotations, namespaceAnnotations map[string]string) (LoadTestType, error) {
+	switch ltType := loadTest.Spec.Type; ltType {
```
vgarvardt, line 33: I do not see any usages of `ltType` var

```suggestion
	switch loadTest.Spec.Type {
```

pkg/backends/fake/fake.go
```diff
@@ -0,0 +1,117 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	loadTestV1 "github.com/hellofresh/kangal/pkg/kubernetes/apis/loadtest/v1"
+)
+
+var (
+	sleepImage = "alpine"
+	imageTag   = "latest"
```
vgarvardt, line 17: I would stick to one of the latest versions, maybe major, but not latest

pkg/backends/fake/fake.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// CheckOrCreateResources check if Fake kubernetes resources have been create, if they have not been create them
+func (c *Fake) CheckOrCreateResources(ctx context.Context) error {
+	// Get the Namespace resource
+	namespace, err := c.kubeClient.CoreV1().Namespaces().Get(ctx, c.loadTest.Status.Namespace, metaV1.GetOptions{})
+	if err != nil {
+		// The LoadTest resource may no longer exist, in which case we stop
+		// processing.
+		if errors.IsNotFound(err) {
```
vgarvardt, line 60: is it fine to ignore all other possible errors here?

pkg/backends/fake/fake.go
```diff
@@ -0,0 +1,117 @@
...
+			return nil
+		}
+	}
+	// Check that we created master job
+	_, err = c.kubeClient.BatchV1().Jobs(namespace.GetName()).Get(ctx, "loadtest-master", metaV1.GetOptions{})
+	if err != nil {
```
vgarvardt, line 66: The fewer nesting levels the better

```
if errors.IsNotFound(err) {
 ...
}

return err
```

pkg/backends/fake/fake.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// CheckOrUpdateStatus check the Fake resources and calculate the current status of the LoadTest from them
+func (c *Fake) CheckOrUpdateStatus(ctx context.Context) error {
+	// Get the Namespace resource
+	namespace, err := c.kubeClient.CoreV1().Namespaces().Get(ctx, c.loadTest.Status.Namespace, metaV1.GetOptions{})
+	if err != nil {
+		// The LoadTest resource may no longer exist, in which case we stop
+		// processing.
+		if errors.IsNotFound(err) {
```
vgarvardt, line 84: is it fine to skip all other errors here?

pkg/controller/jmeter_integration_test.go
```diff
@@ -34,6 +34,7 @@ func TestMain(m *testing.M) {
 }
 
 func TestIntegrationJMeter(t *testing.T) {
+	t.Skip("Skipping JMeter integration tests")
```
vgarvardt, line 37: dev leftover?

pkg/proxy/jmeter_test.go
```diff
@@ -11,13 +11,14 @@ import (
 )
 
 func TestNewJMeterFromHTTPLoadTest(t *testing.T) {
-	r, err := buildMocFormReq(map[string]string{}, "", string(apisLoadTestV1.LoadTestTypeJMeter))
+	ltType := apisLoadTestV1.LoadTestTypeFake
+	r, err := buildMocFormReq(map[string]string{}, "", string(ltType))
 	if err != nil {
```
vgarvardt, line 16:
```suggestion
        require.NoError(t, err)
```

pkg/proxy/proxy_integration_test.go
```diff
@@ -380,11 +357,12 @@ func TestIntegrationGetLoadtestLogs(t *testing.T) {
 
 	for i := 0; i < 5; i++ {
 		//added sleep to wait for kangal controller to create a namespace and start pods
-		testhelper.WaitForResource(5)
+		time.Sleep(4 * time.Second)
```
vgarvardt, line 360: Can this be replaced with the real check of awaited resources with timeout? E.g. check that awaited resources are created/ready every second and stop waiting either when all done, or fail by timeout? This may make tests more stable and in some cases much faster if resources are ready in avg after 2sec and we do not need to wait 2 more.

pkg/backends/fake/fake.go
aleksandrzhiliaev: Not super critical tbh, put I pinned.

pkg/backends/fake/fake.go
aleksandrzhiliaev: Good point, that was like that before but I changed it to return any possible errors to the controller - so it will reschedule the event.

pkg/backends/fake/fake.go
aleksandrzhiliaev: Agree, changed it a bit

pkg/controller/jmeter_integration_test.go
aleksandrzhiliaev: Because of current action limits, we can't run JMeter tests (too heavy), that is why we replaced it by Kangal fake provider tests to test Kangal controller logic only.

pkg/proxy/jmeter_test.go
aleksandrzhiliaev: I didn't really focus on proxy tests in this PR, we have other tickets to rewrite them. Current changes were from the previous branch (from non private repo).

pkg/proxy/proxy_integration_test.go
aleksandrzhiliaev: I didn't really focus on proxy tests in this PR, we have other tickets to rewrite them. Current changes were from the previous branch (from non-private repo).

In this case, we run tests in isolation (kind), so I don't really sure that we will need those timeouts there. As I said we will focus on another ticket.

vgarvardt:
👍

fftorres:
+1

PR #5, opened 2020-09-11
Fix Docker release

aleksandrzhiliaev:
**Description:**
[EES-230] Fix path to the binary in Dockerfile and goreleaser.

[EES-230]: https://hellofresh.atlassian.net/browse/EES-230


iammuho:
+1

PR #7, opened 2020-09-15
Update Kangal chart readme and fix chart path

aleksandrzhiliaev:
**Description**
EES - 132
This PR updates README on Kangal chart by adding instructions on how to fetch Kangal helm package. I also found that we missed one chart path update on goreleaser.


fftorres:
+1

PR #9, opened 2020-09-16
Improve Kangal Helm chart

aleksandrzhiliaev:
**Description**
EES - 3721
The previous version was not passing Helm lint. I made few changes to the chart, enabled Helm lint and put Helm publish the workflow to publish helm chart on master push.

### DIFF
--- a/charts/kangal/Chart.yaml
+++ b/charts/kangal/Chart.yaml
@@ -1,5 +1,14 @@
 apiVersion: v1
-appVersion: "1.0"
-description: A Helm chart for Kubernetes
+description: Run JMeter performance tests in Kubernetes cluster with Kangal
+keywords:
+  - kubernetes
+  - jmeter
+  - performance tests
+  - tests runner
 name: kangal
-version: 1.0.0
+version: 1.0.1
+home: https://github.com/hellofresh/kangal
+icon: https://raw.githubusercontent.com/hellofresh/kangal/master/kangal_logo.svg
+maintainers:
+  - name: Kangal
+    email: kangal@hellofresh.com
--- a/.github/chart_schema.yaml
+++ b/.github/chart_schema.yaml
@@ -1,0 +1,37 @@
+name: str()
+home: str(required=False)
+version: str()
+apiVersion: str()
+appVersion: any(str(), num(), required=False)
+description: str()
+keywords: list(str(), required=False)
+sources: list(str(), required=False)
+maintainers: list(include('maintainer'), required=False)
+dependencies: list(include('dependency'), required=False)
+icon: str(required=False)
+engine: str(required=False)
+condition: str(required=False)
+tags: str(required=False)
+deprecated: bool(required=False)
+kubeVersion: str(required=False)
+annotations: map(str(), str(), required=False)
+type: str(required=False)
+---
+maintainer:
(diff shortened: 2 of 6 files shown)


charts/kangal/Chart.yaml
```diff
@@ -1,5 +1,15 @@
 apiVersion: v1
-appVersion: "1.0"
-description: A Helm chart for Kubernetes
+appVersion: "1.0.0"
```
boekkooi-fresh, line 2: Should this be bumped based on the version being build?

aleksandrzhiliaev: There is no way to bump it automatically with the current setup, so we have to update appVersion manually to trigger new Chart creation.

boekkooi-fresh: So why don't we change the setup and use this Chart as a instead template?

aleksandrzhiliaev: I don't get what you mean, can you please explain?

boekkooi-fresh: I mean that we replace this value a dummy value and then run `sed` and replace it with the version of the app you are releasing. 
That generated version we can then use to publish the chart.

fftorres:
+1

charts/kangal/Chart.yaml
aleksandrzhiliaev: It is two different processes:
 - _Chart release_ happens when you **merge something to master** https://github.com/hellofresh/kangal/pull/9/files#diff-ab2370182407da2a3dc83128b3b23b7fR1 -> it calls GitHub Action https://github.com/helm/chart-releaser-action which checks if there any changes on [charts/kangal](https://github.com/hellofresh/kangal/tree/master/charts/kangal) -> if yes it builds new chart and pushes that to GitHub Releases with name `kangal-1.0.1`
- _Application build release_ happens when you manually create GitHub release with new tag -> it triggers https://github.com/hellofresh/kangal/blob/master/.github/workflows/release.yml workflow and action creates new artifact and Docker image.

It means that we need to bump chart release version every time when we want to update `AppVersion` in Chart.yaml or we don't use `AppVersion` meta which is also an option.

boekkooi-fresh: So based on https://helm.sh/docs/topics/charts/ which states `appVersion: The version of the app that this contains (optional). This needn't be SemVer.` and the default `latest` tag used by https://github.com/hellofresh/kangal/blob/master/charts/kangal/values.yaml#L15 I think there is no reason to have an appVersion in the chart since it's not used.

Regarding the [`version`](https://github.com/hellofresh/kangal/pull/9/files#diff-28cc7b465b89a4a53a8bb23695b49c01R10) it should be incremented every time a new version is release. Do you know if pushing a chart with the same version fails?

aleksandrzhiliaev: Ok, as discussed removed `AppVersion` since we are not using it.
So for version - Helm lint has check that version should be incremented if we touch Helm chart, for example:
```
>>> git show origin/master:charts/kangal/Chart.yaml
Old chart version: 1.0.0
New chart version: 1.0.1
Chart version ok.
```

boekkooi-fresh:
:+1:

PR #11, opened 2020-09-21
Update Kangal cluster role and additionalPrinterColumns

aleksandrzhiliaev:
**Description:**
ES - 4110
Current CRD configuration misses `Age` attribute to show how long the load test resource exists. It also misses a cluster role 
to update LoadTest resource.

```
k get lt
NAME                   TYPE     PHASE      AGE
loadtest-funky-bison   JMeter   finished   8d
```


vgarvardt:
👍

PR #14, opened 2020-09-21
Replace sleep with watch

diegomarangoni:
EES-3884
Using Watch instead of sleep.

### DIFF
--- a/pkg/controller/test_helper.go
+++ b/pkg/controller/test_helper.go
@@ -67,8 +67,7 @@ func readFile(filename string) (string, error) {
 	return str, nil
 }
 
-// WaitLoadtestFunc handles conditional watch events
-func WaitLoadtestFunc(event watch.Event) (bool, error) {
+func waitLoadtestFunc(event watch.Event) (bool, error) {
 	switch event.Type {
 	case watch.Added:
 		return true, nil
@@ -91,12 +90,9 @@ func WaitLoadtest(clientSet clientSetV.Clientset, loadtestName string) error {
 		return err
 	}
 
-	_, err = watchtools.UntilWithoutRetry(ctx, watchObj, WaitLoadtestFunc)
-	if err != nil {
-		return err
-	}
+	_, err = watchtools.UntilWithoutRetry(ctx, watchObj, waitLoadtestFunc)
 
-	return nil
+	return err
 }
 
 // DeleteLoadTest deletes a load test CR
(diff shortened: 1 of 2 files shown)


pkg/controller/test_helper.go
```diff
@@ -65,6 +67,38 @@ func readFile(filename string) (string, error) {
 	return str, nil
 }
 
+// WaitLoadtestFunc handles conditional watch events
+func WaitLoadtestFunc(event watch.Event) (bool, error) {
```
boekkooi-fresh, line 71: Should this be private? since it's only used by `WaitLoadtest`
```suggestion
func waitLoadtestFunc(event watch.Event) (bool, error) {
```

pkg/controller/test_helper.go
```diff
@@ -65,6 +67,38 @@ func readFile(filename string) (string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	_, err = watchtools.UntilWithoutRetry(ctx, watchObj, WaitLoadtestFunc)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
boekkooi-fresh, line 99:
```suggestion
	return err
```

pkg/controller/test_helper.go
```diff
@@ -65,6 +67,38 @@ func readFile(filename string) (string, error) {
...
+		return true, nil
+	case watch.Modified:
+		return true, nil
+	default:
+		return false, nil
+	}
+}
+
+// WaitLoadtest waits until Loadtest resources exists
+func WaitLoadtest(clientSet clientSetV.Clientset, loadtestName string) error {
```
vgarvardt, line 83: if this function is used only in tests (according to file name) then it would be better not to return anything, but pass `t *testing.T` as the first parameter, then call `t.Helper()` in the first line to let testing know this is not really test and fail right here instead of returning an error and checking in the caller if error should always lead to test failure.

diegomarangoni: I think this file, the way it is now, can be easily converted to a generic helper (instead of a test_helper)

boekkooi-fresh: Personally I think the current setup is ok. Mostly because `test_helper.go` is part of the controller package and not postfixed with `_test` causing it it be part of the build of kangal. 
This makes adding the `testing` dependency more of an issue then having something that can test if something was loaded.

It would probably be better to not even have this `test_helper` but using a `internal/test` package which contains something like the `resources.go` or `watch.go` file.
In order to avoid this PR going out of scope I think we can merge and fix the `test_helper` issue another day.

boekkooi-fresh:
:+1:

PR #16, opened 2020-09-22
Improve proxy&controller integration tests

diegomarangoni:
EES-3884

### DIFF
--- a/pkg/controller/controller_integration_test.go
+++ b/pkg/controller/controller_integration_test.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


pkg/controller/controller_integration_test.go
```diff
@@ -18,12 +19,14 @@ func TestIntegrationKangalController(t *testing.T) {
 	if testing.Short() {
 		t.Skip("skipping test in short mode")
 	}
+
 	t.Log()
```
vgarvardt, line 23: log what? =)

diegomarangoni: Forgot to remove :)

vgarvardt:
👍

PR #20, opened 2020-09-24
update charts README.md with crd installation instructions

s-radyuk:
EES-4125
Since we're using Helm v2 and it doesn't support an elegant way of managing CRDs, we update the README with recommendations to apply CRD in the Kubernetes cluster manually.


aleksandrzhiliaev:
+1

aleksandrzhiliaev:
+1

PR #22, opened 2020-09-25
Add support to persist reports using PreSigned URLs

diegomarangoni:
EES-3575

Instead of injecting the AWS into backends containers, this PR adds the ability to inject a PreSigned URL which will allow the backend container to upload reports into it.

Dependent PR:
https://github.com/hellofresh/kangal-jmeter/pull/7


iammuho:
If I can understand from the description, with this PR, we don't have to inject the AWS credentials to the pods? But As I can see from the code we again have to inject the credentials and it will be used to create minio object. The issue, why we need to create presigned URLs but not directly aws s3 commands ? Normally, presigned URls are for customers to upload data to s3 but we have passing credentials already to the pods, so what is the reason to create a minio instance?

diegomarangoni:
> If I can understand from the description, with this PR, we don't have to inject the AWS credentials to the pods?
Correct

> But As I can see from the code we again have to inject the credentials and it will be used to create minio object. The issue, why we need to create presigned URLs but not directly aws s3 commands ?
There is no way to create a aws s3 command and send then to the pod without given the aws credentials as well

> Normally, presigned URls are for customers to upload data to s3 but we have passing credentials already to the pods, so what is the reason to create a minio instance?
The idea is that only `kangal proxy` has access to AWS credentials, as already does for the `/report` endpoint.

iammuho:
> There is no way to create an AWS s3 command and send then to the pod without given the AWS credentials as well

Nope, but you can just send aws credentials and use them in the proxy ?

> The idea is that only Kangal proxy has access to AWS credentials, as already does for the /report endpoint.

Separation of the deployments of the Kangal controller and proxy will also solve this problem without changing the codebase. So you can just pass the creds to the proxy but not controller.

But in general, I'm trying to understand the idea of the using presignedurls with minio rather than simply using aws sdk.

diegomarangoni:
Currently, the only components who actually needs the AWS credentials are `Kangal proxy` and `JMeter backend`
The idea is to have the credentials only on `Kangal proxy` but give a way to the backends (JMeter in this case) to upload their reports

pkg/proxy/server.go
```diff
@@ -74,15 +74,10 @@ func RunServer(ctx context.Context, cfg Config, rr Runner) error {
 	// ---------------------------------------------------------------------- //
 	// LoadTest reports
 	// ---------------------------------------------------------------------- //
-
-	// we need to do redirects because we serve html reports
-	redirectReportHandler := func(w http.ResponseWriter, r *http.Request) {
-		// FileServer removes end slashes from the path and redirects to .. - as a workaround we changed dashboard index.html to main.html
-		url := fmt.Sprintf("%s/main.html", r.URL.Host+r.URL.Path)
+	r.Get("/load-test/{id}/report", func(w http.ResponseWriter, r *http.Request) {
```
aleksandrzhiliaev, line 77: Keep in mind that changing way of exposing reports will require a change in Kangal backend providers too. So this version should pin version of kangal-jmeter at least.

aleksandrzhiliaev (commented):
Can you also add some docs on how reporting flow works in Kangal and what are requirements for new backend providers (aka backend provider should put report tar in root dir)?

pkg/proxy/server.go
diegomarangoni: I change the kangal-jmeter image to be BC

aleksandrzhiliaev:
+1

PR #23, opened 2020-09-25
Adjust jmeter resource requests

s4nji:
This PR re-enables jmeter integration test and adjusts memory request on each container to allow integration test with Github Actions (as there is a 7GB memory limit).

### DIFF
--- a/pkg/backends/jmeter/resources.go
+++ b/pkg/backends/jmeter/resources.go
@@ -204,8 +204,8 @@ func (c *JMeter) NewPod(i int, configMap *coreV1.ConfigMap, podAnnotations map[s
 					},
 					Resources: coreV1.ResourceRequirements{
 						Requests: map[coreV1.ResourceName]resource.Quantity{
-							coreV1.ResourceMemory: resource.MustParse("4Gi"),
-							coreV1.ResourceCPU:    resource.MustParse("1000m"),
+							coreV1.ResourceMemory: resource.MustParse("500Mi"),
+							coreV1.ResourceCPU:    resource.MustParse("250m"),
 						},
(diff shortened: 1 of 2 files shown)


pkg/backends/jmeter/resources.go
```diff
@@ -204,8 +204,8 @@ func (c *JMeter) NewPod(i int, configMap *coreV1.ConfigMap, podAnnotations map[s
 					},
 					Resources: coreV1.ResourceRequirements{
 						Requests: map[coreV1.ResourceName]resource.Quantity{
-							coreV1.ResourceMemory: resource.MustParse("4Gi"),
-							coreV1.ResourceCPU:    resource.MustParse("1000m"),
+							coreV1.ResourceMemory: resource.MustParse("500Mi"),
```
aleksandrzhiliaev, line 207: but on real load tests we need more resources, how will it work?

s4nji: This only adjusts resource requests; if needed it can still get more resources up until the resource limits.

pkg/controller/jmeter_integration_test.go
```diff
@@ -123,9 +122,12 @@ func TestIntegrationJMeter(t *testing.T) {
 
 	t.Run("Checking master pod is created", func(t *testing.T) {
 		var master coreV1.PodList
-		for i := 0; i < 5; i++ {
+		for i := 0; i < 60; i++ {
```
aleksandrzhiliaev, line 125: We might use Watch here too, see https://github.com/hellofresh/kangal/pull/14

s4nji:
Closing this PR in favor of https://github.com/hellofresh/kangal/pull/34.

PR #25, opened 2020-09-28
use net/http context for proxy

iammuho:
When I've started to pprof-ing the Kangal proxy the create handler, I saw, for any request first `net/http` package is creating a context and passing to handler with request and then we are creating another one with timeout. This brings extra alloc. This PR removes extra `context.Background()` but uses the `"net/http".Request.Context`.

net/http context (from godoc)

```
The Request type that is part of the net/http package already has a context attached to it.
As of Go 1.7, Request has a Context function, which returns the request's context.
For incoming server requests, the server cancels the context when the client's connection closes,
when the request is canceled (in HTTP/2), or when the ServeHTTP method returns.
```

Detail: https://golang.org/pkg/net/http/#Request.Context


pkg/proxy/proxy.go
```diff
@@ -63,8 +63,7 @@ func getLoadTestType(r *http.Request) apisLoadTestV1.LoadTestType {
 func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
 	logger := mPkg.GetLogger(r.Context())
 
-	ctx, cancel := context.WithTimeout(context.Background(), loadtest.KubeTimeout)
-	defer cancel()
+	ctx := r.Context()
```
boekkooi-fresh, line 66: Why remove the timeout here and for the other contexts?

I can understand doing the following but removing this deadline seems odd to me,
```suggestion
	ctx, cancel := context.WithTimeout(r.Context(), loadtest.KubeTimeout)
```

iammuho: Yeah, because I've passed this deadlines through `net/http` package. Much clean and easier and performant to be configured from first routine.

boekkooi-fresh: Actually these timeout do slightly different things, since `request != client call` by making them the same you can actually get an cancelled request instead if a nice error response.

aleksandrzhiliaev: That cancel timeout needed to cancel requests to Kube API, not request itself.

iammuho: These are not k8s call but normal API requests from our local or CI.

iammuho: It's the same and with the requests, it's much better. Because currently if the user canceled his requests, in the background, our library is trying to finish the process and unnecessarily uses resources. But with net/http context, if something happens to requests, all the context will be done and stop to trying on the background. I'm not sure if it's designed purposefully to say "yeah, even if something happens, try to finish operations"

boekkooi-fresh: I agree that using the request context is better. 
I do not agree that removing the timeout for `KubeTimeout` is good. This is why I proposed using `ctx, cancel := context.WithTimeout(r.Context(), loadtest.KubeTimeout)`.
This way we allow the request cancellation to propagate and we allow k8s to timeout and still return a nice response.

iammuho: Ah sorry then I missed that! Sure, then I can change it like that

boekkooi-fresh:
:+1:

aleksandrzhiliaev:
+1

PR #26, opened 2020-09-29
checkfiles step for shell scripts

iammuho:
After we move the integration test to its own script, we can add shellcheck to the testing pipeline/workflow to check for shell script errors.

For this step, I've used the `shellcheck` github actions from marketplace: https://github.com/marketplace/actions/shellcheck

Test for intentional mistake: https://github.com/hellofresh/kangal/pull/26/checks?check_run_id=1181914579


aleksandrzhiliaev:
+1

PR #27, opened 2020-09-29
Move LoadTest spec and object creation to own package

aleksandrzhiliaev:
**Description:**
EES - 4160

This PR moves load test resource creation to own `loadtest` package and also cleans proxy side from hardcoded JMeter backend validation process to more generic HTTP validator + load test spec validator.

So the flow will be:

[[HTTP request](https://github.com/hellofresh/kangal/pull/27/files#diff-0530307c704a9dc589d79d6e98bd73aaR83-R84)] -> [[Check body with HTTP validator](https://github.com/hellofresh/kangal/pull/27/files#diff-b4c49c47e87f6d9c9e5d6f1275002b2dR47-R80)] -> [[Make LoadTestSpec](https://github.com/hellofresh/kangal/pull/27/files#diff-a2c7b6b137a14498cd0e5e44b3740a5fR15-R37)] -> [[Build LoadTest Object](https://github.com/hellofresh/kangal/pull/27/files#diff-00316bfb84b07bf4c8a8efefc5db86f0R12-R32)] -> 
              [[Call Kube API to make new LoadTest Resource](https://github.com/hellofresh/kangal/pull/27/files#diff-0530307c704a9dc589d79d6e98bd73aaR98)]

`Make LoadTestSpec` - this part could be moved to specific backends which can implement validation rules based on use case.


pkg/proxy/request.go
```diff
@@ -0,0 +1,138 @@
...
+	backendType     = "type"
+	distributedPods = "distributedPods"
+	testFile        = "testFile"
+	testData        = "testData"
+	envVars         = "envVars"
+	loadTestID      = "id"
+)
+
+//RequestValidator validates request body
+func httpValidator(r *http.Request) url.Values {
```
aleksandrzhiliaev, line 27: The code below is the same as in https://github.com/hellofresh/kangal/blob/master/pkg/proxy/jmeter.go#L58
with few notes:
 - HTTP validator is not checking for different LoadTest ba
ckend types anymore.
 - HTTP validator is not making required distributedPods and testFile - that checks will be happened on domain side, on LoadTestSpec level:
https://github.com/hellofresh/kangal/pull/27/files#diff-a2c7b6b137a14498cd0e5e44b3740a5fR18-R28

That validation part will be also extracted on backend level check - because for example those fields are needed only on one load test backend(JMeter)

pkg/kubernetes/apis/loadtest/v1/loadtest.go
```diff
@@ -0,0 +1,38 @@
...
+		ObjectMeta: metaV1.ObjectMeta{
+			Name:   name,
+			Labels: labels,
+		},
+		Spec: spec,
+		Status: LoadTestStatus{
+			Phase: LoadTestCreating,
+		},
+	}, nil
+}
```
iammuho, line 32: I think we can also write a test for here

pkg/proxy/request.go
```diff
@@ -0,0 +1,138 @@
...
+		Rules:           rules, // rules map
+		RequiredDefault: false, // all the field to be pass the rules,
+	}
+
+	v := govalidator.New(opts)
+	return v.Validate()
+}
+
+// FromHTTPRequestToLoadTestSpec creates a load test spec from HTTP request
+func FromHTTPRequestToLoadTestSpec(r *http.Request, logger *zap.Logger) (apisLoadTestV1.LoadTestSpec, error) {
```
diegomarangoni, line 47: this looks proxy internal function, would be nice to be **not** exported

pkg/proxy/request.go
```diff
@@ -0,0 +1,138 @@
...
+const (
+	backendType     = "type"
+	distributedPods = "distributedPods"
+	testFile        = "testFile"
+	testData        = "testData"
+	envVars         = "envVars"
+	loadTestID      = "id"
+)
+
+//RequestValidator validates request body
```
diegomarangoni, line 26:
```suggestion
//httpValidator validates request body
```

diegomarangoni:
+1

s-radyuk:
+1

PR #28, opened 2020-09-30
overwrite option for the tests

iammuho:
**Problem:**

As a test owner, if I want to update my test or re-create from scratch, I've to delete the test and re-create again. This is extra effort

**Solution:**

I've implemented a `overwrite=true/false` option for the proxy. If `overwrite=true`, proxy first delete the old tests with the hash (same control mechanism for create) then will work as expected. I've cleaned the create function to list and check if there is any `conflict` control but proxy handler.

Open to discussions


pkg/kubernetes/client.go
```diff
@@ -39,24 +38,24 @@ func NewClient(loadTestClient loadTestV1.LoadTestInterface, kubeClient kubernete
 	}
 }
 
-// CreateLoadTest creates new load test from given request data
-func (c *Client) CreateLoadTest(ctx context.Context, loadTest *apisLoadTestV1.LoadTest) (string, error) {
-	c.logger.Debug("Creating load test CR ...")
-
+// ListLoadTests lists the load test from given request data
+func (c *Client) ListLoadTests(ctx context.Context, loadTest *apisLoadTestV1.LoadTest) (*apisLoadTestV1.LoadTestList, error) {
```
aleksandrzhiliaev, line 42: Probably name should be something like `GetLoadTestsByLabel`?
Because I assume if it is a ListLoadTests - it will give me a full list of load tests.

pkg/proxy/proxy.go
```diff
@@ -100,14 +99,32 @@ func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
...
+
+		// If users wants to overwrite
+		if loadTest.Spec.Overwrite == true {
+			for _, item := range labeledLoadTests.Items {
+
+				// Remove the old tests
+				err := p.kubeClient.DeleteLoadTest(ctx, item.Name)
+				if err != nil {
+					logger.Error("Could not delete load test with error:", zap.Error(err))
+					render.Render(w, r, cHttp.ErrResponse(http.StatusBadRequest, err.Error()))
```
aleksandrzhiliaev, line 115: Not sure that this is BadRequest from a user perspective. If we can delete load test should it be 500?

pkg/kubernetes/client.go
iammuho: Yeah, It's giving the full list of load tests, that's why I've iterated. But I can change the name, not a big deal

docs/Kangal-user-flow.md
```diff
@@ -22,6 +22,7 @@ curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
   -F testData=@./artifacts/loadtests/testData.csv \
   -F envVars=@./artifacts/loadtests/envVars.csv \
   -F type=JMeter
+  -F overwrite=true
```
aleksandrzhiliaev, line 25: Please add some description of what does this key mean

pkg/proxy/request.go
```diff
@@ -27,6 +28,7 @@ const (
 func httpValidator(r *http.Request) url.Values {
 	rules := govalidator.MapData{
 		"type":            []string{"required"},
+		"overwrite":       []string{"in:1,True,true,t,T,TRUE,0,False,false,f,F,FALSE"},
```
aleksandrzhiliaev, line 31: if it is boolean and according to https://swagger.io/docs/specification/data-models/data-types/ should only be `true` - `false`.

iammuho: These are accepted values for validator because go parseBool accepts these.

https://golang.org/pkg/strconv/#ParseBool

But if you don't like, I can change

aleksandrzhiliaev: Got it, just want make it consistent to our open API specs

docs/Kangal-user-flow.md
iammuho: Where should I write this description ? There is no description for other files too ? Should I create a table like accepted values etc?

aleksandrzhiliaev: Would be nice to put some description under https://github.com/hellofresh/kangal/blob/master/docs/Kangal-user-flow.md#create section, what do you think?

pkg/proxy/request.go
iammuho: So should I change or leave?

docs/Kangal-user-flow.md
iammuho: Put a table : https://github.com/hellofresh/kangal/blob/dbcf378e0c70e92f427f98ed2913f4fc8b645314/docs/Kangal-user-flow.md

docs/Kangal-user-flow.md
```diff
@@ -22,7 +22,9 @@ curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
   -F testData=@./artifacts/loadtests/testData.csv \
   -F envVars=@./artifacts/loadtests/envVars.csv \
   -F type=JMeter
```
diegomarangoni, line 24:
```suggestion
  -F type=JMeter \
```

openapi.json
```diff
@@ -278,7 +278,10 @@
 					"type": {
 						"type": "string",
 						"enum": ["jMeter", "Fake"]
-					}
+					},
+					"overwrite": {
+						"type": "boolean"
+					},
```
diegomarangoni, line 284:
```suggestion
					}
```

pkg/proxy/proxy.go
```diff
@@ -94,15 +79,47 @@ func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Pushing LoadTest to Kubernetes
-	loadTestName, err := p.kubeClient.CreateLoadTest(ctx, loadTest)
-	if err != nil {
-		if errors.Is(err, os.ErrExist) {
+	// Find the old load test with same data
+	labeledLoadTests, err := p.kubeClient.GetLoadTestsByLabel(ctx, loadTest)
```
diegomarangoni, line 83: You are not checking errors here, is that right?

iammuho: Just changed the lines but there is error checks for this line also below

pkg/proxy/request.go
```diff
@@ -110,6 +118,21 @@ func getFileFromHTTP(r *http.Request, file string) (string, error) {
...
+	if o == "" {
+		return false, nil
+	}
+
+	overwrite, err := strconv.ParseBool(o)
+	if err != nil {
+		return false, err
+	}
+
+	return overwrite, nil
```
diegomarangoni, line 133:
```suggestion
	return strconv.ParseBool(o)
```

iammuho: This is a designed way. Because if the overwrite option is empty, `strconv.ParseBool` gives error but it should be `false` so I've implemented the error handling here.

diegomarangoni: ohh, got it, sorry

pkg/proxy/proxy.go
diegomarangoni: The error is being overwritten, so not tested

pkg/kubernetes/client_test.go
```diff
@@ -100,6 +100,26 @@ func TestCreateLoadTestCRNoLoadTest(t *testing.T) {
 	assert.Error(t, deleteErr)
 }
 
+func TestGetLoadTestsByLabel(t *testing.T) {
```
s-radyuk, line 103: Let's add a negative test as well, to cover both cases.

pkg/proxy/proxy.go
```diff
@@ -94,15 +79,47 @@ func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Pushing LoadTest to Kubernetes
-	loadTestName, err := p.kubeClient.CreateLoadTest(ctx, loadTest)
-	if err != nil {
-		if errors.Is(err, os.ErrExist) {
+	// Find the old load test with same data
```
s-radyuk, line 82:
```suggestion
	// Find the old load test with the same data
```

pkg/kubernetes/client_test.go
iammuho: I couldn't find any way to test negative for this function in fact, because it doesn't do anything but just listing. If you can see `TestGetLoadTest`, there is also no negative test for this.

aleksandrzhiliaev:
+1

PR #29, opened 2020-09-30
Adding unit tests for proxy create method

s-radyuk:
EES-4156
In this PR:
- adding unit tests for proxy Create method
- adding missing tests for client_test.go TestGetLoadTestsByLabel method
- renaming createRequestBody -> createRequestWrapper


pkg/proxy/proxy_test.go
```diff
@@ -155,3 +165,82 @@ func buildMocFormReq(requestFiles map[string]string, distributedPods, ltType str
...
+			"unknown",
+			map[string]string{
+				"testFile": "testdata/valid/loadtest.jmx",
+			},
+			http.StatusBadRequest,
+			"{\"error\":\"invalid LoadTest type\"}\n",
+			"application/json; charset=utf-8",
+		}} {
+		t.Run(tt.name, func(t *testing.T) {
+			var logger = zap.NewNop()
```
vgarvardt, line 219: if you use this

```suggestion
			var logger = zaptest.NewLogger(t)
```

you will get logger output if test fails

pkg/proxy/proxy.go
```diff
@@ -30,17 +30,21 @@ var (
...
 
+type loadTestSpecCreator func(*http.Request, *zap.Logger) (apisLoadTestV1.LoadTestSpec, error)
+
 // Proxy handler
 type Proxy struct {
-	maxLoadTestsRun int
-	kubeClient      *kube.Client
+	maxLoadTestsRun     int
+	kubeClient          *kube.Client
+	httpToSpecConverter loadTestSpecCreator
```
aleksandrzhiliaev, line 39: No need to inject this internal function here

pkg/proxy/proxy.go
```diff
@@ -30,17 +30,21 @@ var (
 	ErrFileToStringEmpty error = errors.New("file is empty")
 )
 
+type loadTestSpecCreator func(*http.Request, *zap.Logger) (apisLoadTestV1.LoadTestSpec, error)
```
aleksandrzhiliaev, line 33:
```suggestion
var fromHTTPToSpecCreatorFn = fromHTTPRequestToLoadTestSpec
```
then use it as I showed below

pkg/proxy/proxy.go
```diff
@@ -66,7 +70,7 @@ func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
 	defer cancel()
 
 	// Making valid LoadTestSpec based on HTTP request
-	ltSpec, err := fromHTTPRequestToLoadTestSpec(r, logger)
+	ltSpec, err := p.httpToSpecConverter(r, logger)
```
aleksandrzhiliaev, line 73:
```suggestion
	ltSpec, err := fromHTTPToSpecCreatorFn(r, logger)
```

pkg/proxy/proxy.go
vgarvardt: package-level vars is bad idea in general, explicit deps injection is much more straightforward

aleksandrzhiliaev: It doesn't have any shared state to be bad, just another way of dealing with dependencies.

vgarvardt: Just a legal way of monkey-patching for tests 😜

pkg/proxy/proxy.go
s-radyuk: I find the current implementation more transparent. It allows me to mock this dependency in tests in a straightforward way.

aleksandrzhiliaev:
+1

PR #33, opened 2020-10-02
Setup CodeQL analysis according to GH wizard

RTodorov:
This is the standard codeql configuration file proposed by the Github UI wizard


fftorres:
+1

PR #34, opened 2020-10-02
Add config to adjust JMeter pod resource limits and requests

s4nji:
This PR:
- Adds environment variables to adjust JMeter master and worker pod resource (memory and cpu) limits and request
- Lowers cpu and memory resource requests for both master and worker in integration test
- (Re-)enables jmeter integration-test

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -54,14 +54,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@master
 
-      - name: Run Unit tests
-        run: |
-          make test-unit
-
       - name: Setup Golang
         uses: actions/setup-go@v2
         with:
           go-version: '^1.15'
+
+      - name: Run Unit tests
+        run: |
+          make test-unit
 
       - name: Setup Kind
         uses: engineerd/setup-kind@v0.4.0
--- a/pkg/backends/jmeter/config.go
+++ b/pkg/backends/jmeter/config.go
@@ -2,12 +2,12 @@ package jmeter
 
 // Config specific to JMeter backend
 type Config struct {
-	MasterCPULimits      string `envconfig:"JMETER_MASTER_CPU_LIMITS" default:"2000m"`
-	MasterCPURequests    string `envconfig:"JMETER_MASTER_CPU_REQUESTS" default:"1000m"`
-	MasterMemoryLimits   string `envconfig:"JMETER_MASTER_MEMORY_LIMITS" default:"4Gi"`
-	MasterMemoryRequests string `envconfig:"JMETER_MASTER_MEMORY_REQUESTS" default:"4Gi"`
-	WorkerCPULimits      string `envconfig:"JMETER_WORKER_CPU_LIMITS" default:"2000m"`
-	WorkerCPURequests    string `envconfig:"JMETER_WORKER_CPU_REQUESTS" default:"1000m"`
-	WorkerMemoryLimits   string `envconfig:"JMETER_WORKER_MEMORY_LIMITS" default:"4Gi"`
-	WorkerMemoryRequests string `envconfig:"JMETER_WORKER_MEMORY_REQUESTS" default:"4Gi"`
+	MasterCPULimits      string `envconfig:"JMETER_MASTER_CPU_LIMITS"`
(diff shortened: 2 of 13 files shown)


.github/workflows/testing.yml
```diff
@@ -54,15 +54,15 @@ jobs:
...
-          make test-unit
-
       - name: Setup Golang
         uses: actions/setup-go@v2
         with:
           go-version: '^1.15'
 
+      - name: Run Unit tests
+        run: |
+          make test-unit
```
aleksandrzhiliaev, line 64: How did it work before without setting up Golang? :-D

pkg/controller/jmeter_integration_test.go
```diff
@@ -123,9 +122,12 @@ func TestIntegrationJMeter(t *testing.T) {
 
 	t.Run("Checking master pod is created", func(t *testing.T) {
 		var master coreV1.PodList
-		for i := 0; i < 5; i++ {
+		for i := 0; i < 60; i++ {
```
aleksandrzhiliaev, line 125: can we add `watch` there instead?

pkg/controller/loadtest.go
```diff
@@ -301,7 +301,7 @@ func (c *Controller) syncHandler(key string) error {
 	loadTest := loadTestFromCache.DeepCopy()
 	defer c.updateLoadTestStatus(ctx, loadTest)
 
-	backend, err := backends.NewLoadTest(loadTest, c.kubeClientSet, c.kangalClientSet, c.logger, c.namespacesLister, c.cfg.Report, c.cfg.PodAnnotations, c.cfg.NamespaceAnnotations)
+	backend, err := backends.NewLoadTest(loadTest, c.kubeClientSet, c.kangalClientSet, c.logger, c.namespacesLister, c.cfg.Report, c.cfg.PodAnnotations, c.cfg.NamespaceAnnotations, c.cfg.Backends)
```
aleksandrzhiliaev, line 304: Would be nice to cover all Codecov comments by some simple tests.

.github/workflows/testing.yml
s4nji: Good question; it was broken only locally when I ran it with act.

pkg/controller/loadtest.go
s4nji: Added via 55b8af1.

pkg/controller/jmeter_integration_test.go
s4nji: Done via 83f4a29.

pkg/backends/jmeter/config.go
```diff
@@ -0,0 +1,13 @@
+package jmeter
+
+// Config specific to JMeter backend
+type Config struct {
+	MasterCPULimits      string `envconfig:"JMETER_MASTER_CPU_LIMITS" default:"2000m"`
```
aleksandrzhiliaev, line 5: Should it be another way around?
So we have small defaults and we can override those values for production load testing.

s4nji: I think the current default numbers are fair; we only adjust the numbers in integration test because github actions (free) runner is really limited in resource.  
In production you do want somewhat high resource requests and limits, otherwise your load generator might "fail" before the service does, and you get innacurate results.

aleksandrzhiliaev: I don't agree with that tbh. Can we even leave that empty here and override only if it is needed?
Because for example I want to run Kangal locally but I can't because it requires high specs by default. But for my small test my local laptop specs should be fine - so not really needed to set any limits at all.

s4nji: Resource limits and requests are no longer required as of 1fb2cab.

pkg/backends/jmeter/config.go
```diff
@@ -0,0 +1,13 @@
+package jmeter
+
+// Config specific to JMeter backend
+type Config struct {
+	MasterCPULimits      string `envconfig:"JMETER_MASTER_CPU_LIMITS"`
```
aleksandrzhiliaev, line 5: We should put some docs about that vars and why they are needed

aleksandrzhiliaev:
+1

diegomarangoni:
+1

PR #37, opened 2020-10-06
Implements Locust

diegomarangoni:
EES-4138


pkg/backends/jmeter/resources_test.go
```diff
@@ -66,32 +66,6 @@ func TestSplitTestDataTrimComma(t *testing.T) {
...
-func TestReadSecretInvalid(t *testing.T) {
-	teststring := "aaa:1\nbbb;2\nccc;3\n"
-	expectedError := os.ErrInvalid
-
-	_, err := readSecret(teststring, logger)
-	assert.Error(t, err)
-	assert.Equal(t, expectedError, err)
-}
-
-func TestReadSecretEmpty(t *testing.T) {
```
diegomarangoni, line 87: Moved those tests to `helper` package

pkg/backends/jmeter/resources.go
```diff
@@ -416,29 +406,6 @@ func splitTestData(testdata string, n int, logger *zap.Logger) ([][][]string, er
...
-		}
-		if len(line) != 2 {
-			logger.Error("Invalid secret format in envVars file")
-			return nil, os.ErrInvalid
-		}
-		m[line[0]] = line[1]
-	}
-	return m, nil
-}
-
```
diegomarangoni, line 441: Move this function to a dedicated package called `helper`, so Locust can make use of it too.

pkg/backends/jmeter/resources.go
```diff
@@ -449,31 +416,3 @@ func getNamespaceFromLoadTestName(loadTestName string, logger *zap.Logger) (newN
 	newNamespaceName = nsName[loadTestNameLength-2] + "-" + nsName[loadTestNameLength-1]
 	return newNamespaceName, nil
 }
-
-// buildResourceRequirements creates ResourceRequirements that allows not all values to be specified
-// This is necessary because setting a resource requirement with value 0 produces a different behavior
-func buildResourceRequirements(cpuLimit, cpuRequest, memLimit, memRequest string) coreV1.ResourceRequirements {
```
diegomarangoni, line 455: Also moved this function to the same `helper` package to use with Locust

pkg/proxy/request.go
```diff
@@ -30,9 +33,11 @@ func httpValidator(r *http.Request) url.Values {
 		"type":            []string{"required"},
 		"overwrite":       []string{"in:1,True,true,t,T,TRUE,0,False,false,f,F,FALSE"},
 		"distributedPods": []string{"numeric_between:1,"},
-		"file:testFile":   []string{"ext:jmx"},
+		"file:testFile":   []string{"ext:jmx,py"},
```
s-radyuk, line 36: This way we can create Locust test with .jmx file instead of .py because both file types are valid. But it doesn't make sense. And the test will fail only after creating all the resources and failing on start. 
Let's maybe add validation of file type to fromHTTPRequestToLoadTestSpec?

pkg/backends/locust/locust.go
```diff
@@ -0,0 +1,211 @@
...
+
+	"github.com/hellofresh/kangal/pkg/core/helper"
+	loadTestV1 "github.com/hellofresh/kangal/pkg/kubernetes/apis/loadtest/v1"
+	clientSetV "github.com/hellofresh/kangal/pkg/kubernetes/generated/clientset/versioned"
+	"go.uber.org/zap"
+)
+
+var (
+	defaultImage    = "locustio/locust"
+	defaultImageTag = "latest"
```
s-radyuk, line 21: Should we pin the version here?

pkg/backends/locust/locust.go
```diff
@@ -0,0 +1,211 @@
...
+	loadTest           *loadTestV1.LoadTest
+	logger             *zap.Logger
+	reportPreSignedURL *url.URL
+	masterResources    helper.Resources
+	workerResources    helper.Resources
+	podAnnotations     map[string]string
+}
+
+// SetDefaults mutates the LoadTest object to add default values to empty fields
+func (c *Locust) SetDefaults() error {
```
s-radyuk, line 37: This is a duplicated code, would it be better to extract it to the helper?

README.md
```diff
@@ -47,6 +48,15 @@ Kangal application contains two main parts:
...
 Kangal also uses S3 compatible storage to save test reports. 
 
+## Supported backends
+Currently, there are the following load generator types implemented for Kangal:
+
+- **Fake**
+- [**JMeter**](https://jmeter.apache.org/)
+- [**Locust**](https://locust.io/)
+
+Read more about each of then at [docs/README.md](docs/README.md).
```
s-radyuk, line 58:
```suggestion
Read more about each of them in [docs/README.md](docs/README.md).
```

pkg/kubernetes/apis/loadtest/v1/types.go
```diff
@@ -42,14 +44,16 @@ type LoadTestPodsStatus struct {
...
-	EnvVars         string       `json:"envVars,omitempty"`
+	Type            LoadTestType  `json:"type"`
+	Overwrite       bool          `json:"overwrite"`
+	MasterConfig    ImageDetails  `json:"masterConfig"`
+	WorkerConfig    ImageDetails  `json:"workerConfig"`
+	DistributedPods *int32        `json:"distributedPods"`
+	TestFile        string        `json:"testFile"`
+	TestData        string        `json:"testData,omitempty"`
+	EnvVars         string        `json:"envVars,omitempty"`
+	TargetURL       string        `json:"targetURL,omitempty"`
```
s-radyuk, line 55: Are target url and duration allowed to be empty? As I see locust test doesn't do anything without these params specified. 
I think we shouldn't create a loadtest spec for locust type if those params are empty.

docs/locust/README.md
```diff
@@ -0,0 +1,69 @@
...
+LOCUST_MASTER_CPU_REQUESTS
+LOCUST_MASTER_MEMORY_LIMITS
+LOCUST_MASTER_MEMORY_REQUESTS
+LOCUST_WORKER_CPU_LIMITS
+LOCUST_WORKER_CPU_REQUESTS
+LOCUST_WORKER_MEMORY_LIMITS
+LOCUST_WORKER_MEMORY_REQUESTS
+```
+
+You have to specify this variables on Kangal Controller, read more at [charts/kangal/README.md](/charts/kangal/README.md#kangal-controller-locust-specific).
```
s-radyuk, line 34:
```suggestion
You have to specify these variables on Kangal Controller, read more at [charts/kangal/README.md](/charts/kangal/README.md#kangal-controller-locust-specific).
```

pkg/proxy/request.go
diegomarangoni: It would not fail, as after reading the file it will be handled as string the whole time.
I would prefer to remove file extension validation at all, as we are not actually protecting anything.
And then, on a different PR, implement mimetype validation and some sort of per backend rules.

pkg/backends/locust/locust.go
diegomarangoni: I don't think we should PIN here, but in the helm chart definitely, I will change that

pkg/proxy/request_test.go
```diff
@@ -325,3 +329,20 @@ func TestCheckLoadTestSpec(t *testing.T) {
 	assert.NotEmpty(t, lt.Name)
 	assert.NotEmpty(t, lt.Labels)
 }
+
+func TestGetDuration(t *testing.T) {
```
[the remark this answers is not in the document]
diegomarangoni: That is covered by golang package, but I will add one anyways :)

docs/locust/README.md
```diff
@@ -0,0 +1,69 @@
+# Locust
+
+## Table of content
+- [Configuring Locust resource requirements](#configuring-locust-resource-requirements)
+- [Writing tests](#writing-tests)
+- [Reporting](#reporting)
+
+Locust is one of the load generators used in Kangal. An easy to use, scriptable and scalable performance testing tool. You define the behaviour of your users in regular Python code, instead of using a clunky UI or domain specific language. This makes Locust infinitely expandable and very developer friendly.
```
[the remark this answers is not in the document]
diegomarangoni: Sure, I will remove, I copied from Locust official repo

pkg/backends/locust/locust.go
diegomarangoni: Done!

pkg/backends/locust/locust.go
diegomarangoni: I had to change a little bit, so now it's different

pkg/proxy/request_test.go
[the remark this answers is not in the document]
diegomarangoni: Done!

docs/locust/README.md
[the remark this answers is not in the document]
diegomarangoni: Done!

pkg/kubernetes/apis/loadtest/v1/types.go
diegomarangoni: They both can be empty yes.
For targetURL, you can use full URLs inside the test script.
For duration, Locust allows you to run tests forever if you want to.

docs/locust/README.md
```diff
@@ -0,0 +1,128 @@
+# Locust
+
+## Table of content
+- [How it works](#how-it-works)
+- [Configuring Locust resource requirements](#configuring-locust-resource-requirements)
+- [Writing tests](#writing-tests)
+- [Reporting](#reporting)
+
+Locust is one of the load generators used in Kangal and uses the official docker image [locustio/locust](https://hub.docker.com/r/locustio/locust).
```
s-radyuk, line 9:
```suggestion
Locust is one of the load generators implemented in Kangal. It uses the official docker image [locustio/locust](https://hub.docker.com/r/locustio/locust).
```

s-radyuk:
+1

pkg/core/helper/errors.go
```diff
@@ -0,0 +1,8 @@
+package helper
```
aleksandrzhiliaev, line 1: This is not really go-way of making helers, utils or other package names.

All code belongs to one domain should be in the same package.

pkg/core/helper/helper.go
```diff
@@ -0,0 +1,66 @@
...
+	"encoding/csv"
+	"io"
+	"strings"
+
+	coreV1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// Resources contains resources limits/requests
+type Resources struct {
```
aleksandrzhiliaev, line 13: In my opinion, it should go to the `pkg/core/resource package`

pkg/core/helper/helper.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	return coreV1.ResourceRequirements{
+		Limits:   limits,
+		Requests: requests,
+	}
+}
+
+// ReadEnvs reads data from csv file to save it as a map for creating a secret
+func ReadEnvs(envVars string) (map[string]string, error) {
```
aleksandrzhiliaev, line 49: In my opinion it should go to `pkg/core/csv` package

pkg/proxy/request.go
aleksandrzhiliaev: Agree with @diegomarangoni  about removing file validation, doesn't make sense. All validation logic should happen on the spec side of the backend.

pkg/core/helper/errors.go
diegomarangoni: Agreed, don't know why I did this way.
I opened a ticket EES-4269

pkg/backends/locust/spec.go
```diff
@@ -0,0 +1,27 @@
...
+//BuildLoadTestSpec validates input and returns valid LoadTestSpec
+func BuildLoadTestSpec(overwrite bool, distributedPods int32, testFileStr, envVarsStr, targetURL string, duration time.Duration) (loadTestV1.LoadTestSpec, error) {
+	lt := loadTestV1.LoadTestSpec{}
+	if distributedPods <= int32(0) {
+		return lt, ErrRequireMinOneDistributedPod
+	}
+	if testFileStr == "" {
+		return lt, ErrRequireTestFile
+	}
+	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeLocust, overwrite, distributedPods, testFileStr, "", envVarsStr, loadTestV1.ImageDetails{}, loadTestV1.ImageDetails{}, targetURL, duration), nil
```
aleksandrzhiliaev, line 26:
```golang
loadTestV1.ImageDetails{}, loadTestV1.ImageDetails{}
```
Should have defined image details, see:
https://github.com/hellofresh/kangal/blob/master/pkg/backends/fake/spec.go#L12
https://github.com/hellofresh/kangal/blob/master/pkg/backends/jmeter/spec.go#L27

diegomarangoni: Instead of inserting into the Spec and persisting it, I'm using global default.
This way it will works using `kubectl apply` with yaml file even in case the image fields are left blank.

See:
https://github.com/hellofresh/kangal/blob/master/pkg/backends/locust/locust.go#L39
https://github.com/hellofresh/kangal/blob/master/pkg/backends/locust/locust.go#L193

aleksandrzhiliaev: That was a hack before because it didn't persist specs properly, right now we build fully contained specs in the Proxy side.

diegomarangoni: IMO the default image has to be set on the Controller, and if no image is set on the Spec, the Controller defaults should be use.
This keeps Proxy optional.

If we let the Proxy populates the Spec with image defaults, the ones set in the Controller will never be use.
And Controller would still have to check if Spec image is empty because LoadTest resource can be `kubectl` applied.

aleksandrzhiliaev: Yes, that is an idea to build full spec LoadTest in the proxy side, so controller only takes specs and puts LoadTest to the desired state. So LoadTest spec will be a kinda immutable struct, the controller only responsible on changing statuses.

As I said that SetDefaults was a hack on the controller side because proxy didn't populate load test spec properly.

aleksandrzhiliaev: Making `kubectl apply` for such complex CRD also not a good way (that is why we made easy for humans API on top) but for sure possible. In my opinion, if something is invalid in LoadTest spec controller should set Error state and not making any changes inside a spec.

diegomarangoni: This is the same discussion at https://github.com/hellofresh/kangal/pull/42

I think I got your point, you are saying that Spec images are mandatory and if Controller find an empty string it should throw and error?

aleksandrzhiliaev: That was an initial idea but it didn't work because we had a mistake on our crd.yaml and we made this hack to populate that defaults every time (basically it was never persisting those values, so each sync it was setting that defaults :-D)

PR #39, opened 2020-10-08
Add tags to load tests

nhatthm:
## Motivation

As an end user, I have a need of tagging my load tests and be able to retrieve the list of my tests. There are a couple of use cases:

- I could tag the test with my team name and other team members could easily find out what tests we have
- I could list out all the finished or running tests of my team.

## Changes

In the `ObjectMeta` we will inject the labels so that we can use `labelSelector` to retrieve them back. The label will in format `test-tag-[LABEL]: [VALUE]`

_Creation_

There will be one more param `tags`. For example: `tags=department:platform,team=kangal`

We will create an object with labels

```
test-file-hash: <hash>
test-tag-department: platform
test-tag-team: kangal
```

_List_

The request will be like `/load-test?phase=running&tags=department:platform,team:kangal`

We will use `labelSelector` to find all matched objects, then filter again with `phase`

## Further Benefits

We can integrate with other automation system with list request for
- Download and post report to s3
- Scheduled to delete finished tests.

### DIFF
--- a/docs/user-flow.md
+++ b/docs/user-flow.md
@@ -35,6 +35,20 @@ curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
   -F overwrite=true
 ```
 
+You can also tag the load test so that you can find them later, the format is `tag1:value1,tag2:value2`
+
+```bash
+curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
+  -H 'Content-Type: multipart/form-data' \
+  -F distributedPods=1 \
+  -F testFile=@examples/constant_load.jmx \
+  -F testData=@artifacts/loadtests/testData.csv \
+  -F envVars=@artifacts/loadtests/envVars.csv \
+  -F type=JMeter \
+  -F tags=tag1:value1,tag2:value2 \
+  -F overwrite=true
+```
+
 ## Check 
 Check the status of the load test.
 
@@ -66,4 +80,24 @@ Delete your finished load test.
 
 ```
 curl -X DELETE http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-name
+```
+
+## List
+
+You can find out all the load tests
+
+```bash
+curl http://${KANGAL_PROXY_ADDRESS}/load-test
+```
+
+You can filter by `tags`
+
+```bash
+curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?tags=tag1:value1'
+```
+
+And limit your search
+
+```bash
+curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?tags=tag1:value1&limit=10'
 ```
--- a/openapi.json
+++ b/openapi.json
@@ -14,6 +14,62 @@
 	}],
 	"paths": {
 		"/load-test": {
+			"get": {
+				"tags": ["load-tests"],
+				"summary": "List all the load tests",
+				"operationId": "listLoadTest",
+				"parameters": [
+					{
+						"name": "tags",
+						"in": "query",
+						"description": "Filter the result by tags, value is in format: tag1:value1,tag2:value2",
+						"schema": {
+							"type": "string"
+						},
+						"example": "department:platform,team:kangal"
+					},
+					{
+						"name": "limit",
+						"in": "query",
+						"description": "Limit the result when querying on a large cluster",
+						"schema": {
+							"type": "integer"
+						},
+						"example": "500"
+					},
+					{
+						"name": "continue",
+						"in": "query",
+						"description": "Continue the next page when querying on a large cluster",
+						"schema": {
+							"type": "integer"
+						},
+						"example": "ENCODED_CONTINUE_TOKEN"
+					}
+				],
+				"responses": {
+					"200": {
+						"description": "Expected response to a valid request",
+						"content": {
+							"application/json": {
+								"schema": {
+									"$ref": "#/components/schemas/LoadTestStatusPage"
+								}
+							}
+						}
+					},
+					"default": {
+						"description": "unexpected error",
+						"content": {
+							"application/json": {
+								"schema": {
+									"$ref": "#/components/schemas/Error"
+								}
+							}
+						}
+					}
+				}
+			},
 			"post": {
 				"tags": ["load-tests"],
 				"summary": "Create a new loadTest",
@@ -290,6 +346,27 @@
 					}
 				}
 			},
+			"LoadTestStatusPage": {
+				"type": "object",
+				"properties": {
+					"limit": {
+						"type": "integer"
+					},
+					"continue": {
+						"type": "string"
+					},
+					"remain": {
+						"type": "integer",
+						"nullable": true
+					},
+					"items": {
+						"type": "array",
+						"items": {
+							"$ref": "#/components/schemas/LoadTestStatus"
+						}
+					}
+				}
+			},
 			"LoadTestStatus": {
 				"type": "object",
 				"properties": {
@@ -303,6 +380,12 @@
 					"phase": {
 						"type": "string",
 						"enum": ["creating", "starting", "running", "finished", "errored"]
+					},
+					"tags": {
+						"type": "object",
+						"additionalProperties": {
+							"type": "string"
+						}
 					},
 					"hasEnvVars": {
 						"type": "boolean"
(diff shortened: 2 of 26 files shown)


s-radyuk:
Hello Nhat! 
First of all, thank you very much for your attention to Kangal and your initiative with this proposal! We appreciate it!

I general I find the idea of labeling load tests quite useful. But I want to draw your attention to the fact that finished and errored load tests are automatically deleted from the cluster after 1 hour. Users will not be able to list them to find out what tests your team has. Load tests are temporal and aren't designed to live in the cluster for a long time. So you can list only those that are currently running. And we expect the team to be aware of what they have running at the moment because it can affect the productivity of the whole system.

openapi.json
```diff
@@ -14,6 +14,60 @@
 	}],
 	"paths": {
 		"/load-test": {
+			"get": {
```
s-radyuk, line 17: Please also add an example of usage together with an explanation to the doc [docs/Kangal-user-flow.md](docs/Kangal-user-flow.md)

nhatthm:
> I want to draw your attention to the fact that finished and errored load tests are automatically deleted from the cluster after 1 hour. Users will not be able to list them to find out what tests your team has.

Thanks, I didn't know this

pkg/kubernetes/apis/loadtest/v1/spec.go
```diff
@@ -27,9 +33,16 @@ func BuildLoadTestSpec(loadTestType LoadTestType, overwrite bool, distributedPod
 		return lt, ErrRequireTestFile
 	}
 
+	for _, value := range tags {
+		if len(value) > maxTagLength {
```
boekkooi-fresh, line 37: Should we allow empty tags? Or would it be better to filter them out?

pkg/kubernetes/client.go
```diff
@@ -92,6 +102,40 @@ func (c *Client) GetLoadTest(ctx context.Context, loadTest string) (*apisLoadTes
 	return result, nil
 }
 
+// ListLoadTest returns list of load tests.
+func (c *Client) ListLoadTest(ctx context.Context, opt ListOptions) ([]apisLoadTestV1.LoadTest, error) {
+	k8sOpt := metaV1.ListOptions{}
```
boekkooi-fresh, line 107: I think we should introduce pagination here to avoid a possibly overloading the service if there are a lot of load tests.
See Limit and Continue under [ListOptions](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions)

pkg/proxy/proxy.go
```diff
@@ -50,18 +51,90 @@ func NewProxy(maxLoadTestsRun int, kubeClient *kube.Client, specCreator loadTest
...
 }
 
+// List lists all the load tests.
+func (p *Proxy) List(w http.ResponseWriter, r *http.Request) {
+	logger := mPkg.GetLogger(r.Context())
+
+	ctx, cancel := context.WithTimeout(r.Context(), loadtest.KubeTimeout)
+	defer cancel()
+
+	opt := kube.ListOptions{}
```
boekkooi-fresh, line 74: I think we should introduce pagination here to avoid a possibly overloading the service if there are a lot of load tests.
See Limit and Continue under [ListOptions](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions)

pkg/proxy/proxy.go
```diff
@@ -50,18 +51,90 @@ func NewProxy(maxLoadTestsRun int, kubeClient *kube.Client, specCreator loadTest
...
+
+			value := strings.TrimSpace(parts[1])
+
+			if len(value) != 2 {
+				continue
+			}
+
+			opt.Tags[label] = value
+		}
+	}
```
boekkooi-fresh, line 109: Can we extract this into a func please?

pkg/proxy/proxy_test.go
```diff
@@ -595,9 +726,45 @@ type specData struct {
...
+		if len(value) == 0 {
+			continue
+		}
+
+		if len(value) > maxTagLength {
+			continue
+		}
+
+		tags[label] = value
+	}
```
boekkooi-fresh, line 763: Do we really need to duplicate this logic here? If we would extract https://github.com/hellofresh/kangal/pull/39/files#diff-0530307c704a9dc589d79d6e98bd73aaR85-R108 that func could just be called.

pkg/proxy/request.go
```diff
@@ -159,3 +180,50 @@ func fileToString(f io.ReadCloser) (string, error) {
...
+
+		if len(value) == 0 {
+			return nil, ErrTagMissingValue
+		}
+
+		if len(value) > maxTagLength {
+			return nil, ErrTagValueMaxLengthExceeded
+		}
+
+		tags[label] = value
```
boekkooi-fresh, line 221: This looks very similar to https://github.com/hellofresh/kangal/pull/39/files#diff-0530307c704a9dc589d79d6e98bd73aaR85-R108 and https://github.com/hellofresh/kangal/pull/39/files#diff-047152d6343305cf4a2ae0cf8a7611b3R735-R763 
Is there a reason why we can't fine a way to avoid this duplication? Maybe introduce a `type Tags map[string]string` and a constructor method for it?

nhatthm: This one is for parsing the input from POST request, we're forming the tags and validate it at the same time

pkg/proxy/proxy_test.go
nhatthm: This is a mock parser, the intention is to ignore the error because we don't need it. But we can return error, no problem with that. So I can merge with the one from https://github.com/hellofresh/kangal/pull/39#discussion_r502276632

pkg/proxy/proxy.go
nhatthm: This one is for parsing the tags to filter so we don't need the validation

pkg/kubernetes/apis/loadtest/v1/spec.go
nhatthm: The input is validated at the request level (not empty, not longer than 63 chars) so I don't think we need to check the empty again coz k8s accepts empty label.

docs/user-flow.md
```diff
@@ -25,6 +25,21 @@ curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
...
+```bash
+curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
+  -H 'Content-Type: multipart/form-data' \
+  -F distributedPods=1 \
+  -F testFile=@examples/constant_load.jmx \
+  -F testData=@artifacts/loadtests/testData.csv \
+  -F envVars=@artifacts/loadtests/envVars.csv \
+  -F type=JMeter \
+   # Specify your tags \
+  -F tags=tag1:value1,tag2:value2 \
```
diegomarangoni, line 39:
```suggestion
  -F tags=tag1:value1,tag2:value2 \
```

Copying and pasting will make the command break with this comment

diegomarangoni:
+1

pkg/kubernetes/apis/loadtest/v1/spec.go
boekkooi-fresh, line 37: Do we still need this check? Since a type of `LoadTestTags` should be created by [`parseTags`](https://github.com/hellofresh/kangal/pull/39/files#diff-b4c49c47e87f6d9c9e5d6f1275002b2dR223).

pkg/proxy/request.go
```diff
@@ -159,3 +215,48 @@ func fileToString(f io.ReadCloser) (string, error) {
 
 	return s, nil
 }
+
+func getTags(r *http.Request) (apisLoadTestV1.LoadTestTags, error) {
+	return parseTags(r.FormValue(tags))
+}
+
+func parseTags(tagsStr string) (apisLoadTestV1.LoadTestTags, error) {
```
boekkooi-fresh, line 223: Would it make sense to make this func part of `apisLoadTestV1` as `LoadTestTagsFromString` for example?

pkg/kubernetes/apis/loadtest/v1/spec.go
nhatthm: imo, we should keep the validation because it's k8s thingy but I can remove it if you think it's too much.

openapi.json
```diff
@@ -14,6 +14,60 @@
...
+					},
+					{
+						"name": "tags",
+						"in": "query",
+						"description": "Filter the result by tags, value is in format: tag1:value1,tag2:value2",
+						"schema": {
+							"type": "string"
+						},
+						"example": "department:platform,team:kangal"
+					}
```
diegomarangoni, line 46: It's not missing the `limit` field here?

nhatthm: sharp eyes 👍

nhatthm:
I have to remove `phase` filter for `/load-test` API because it's not supported by k8s

Ref:
- https://github.com/kubernetes/kubernetes/issues/53459
- https://github.com/kubernetes/kubernetes/issues/51046

diegomarangoni:
+1

diegomarangoni:
+1

PR #41, opened 2020-10-14
Add new Proxy endpoint for report persistance

diegomarangoni:
EES-4260

Using Presigned URL's to persist reports does not works well when using IAM roles to obtain credentials.
That's because even if you issue a URL for a long period, you still limited to the credentials expire time, which is 15 minutes, and no way to control that.

This PR adds a new endpoint to Kangal Controller (PUT /load-test/{id}/report) which generates a fresh Presigned URL and stream the request content to it.

### DIFF
--- a/pkg/proxy/server.go
+++ b/pkg/proxy/server.go
@@ -84,7 +84,7 @@ func RunServer(ctx context.Context, cfg Config, rr Runner) error {
 		http.Redirect(w, r, url, http.StatusMovedPermanently)
 	})
 	r.Get("/load-test/{id}/report/*", report.ShowHandler())
-	r.Put("/load-test/{id}/report", report.PersistHandler(rr.KubeClient))
+	r.Put("/load-test/{id}/report", report.PersistHandler(rr.KubeClient, rr.Logger))
 
 	address := fmt.Sprintf(":%d", cfg.HTTPPort)
 	rr.Logger.Info("Running HTTP server...", zap.String("address", address))
--- a/pkg/proxy/server.go
+++ b/pkg/proxy/server.go
@@ -79,6 +79,7 @@ func RunServer(ctx context.Context, cfg Config, rr Runner) error {
 		http.Redirect(w, r, url, http.StatusMovedPermanently)
 	})
 	r.Get("/load-test/{id}/report/*", report.ShowHandler())
+	r.Put("/load-test/{id}/report", report.PersistHandler(rr.KubeClient))
 
 	address := fmt.Sprintf(":%d", cfg.HTTPPort)
 	rr.Logger.Info("Running HTTP server...", zap.String("address", address))
--- a/pkg/report/handler.go
+++ b/pkg/report/handler.go
@@ -2,10 +2,15 @@ package report
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 
 	"github.com/go-chi/chi"
+
+	kube "github.com/hellofresh/kangal/pkg/kubernetes"
 )
+
+var httpClient = &http.Client{}
 
 //ShowHandler method returns response from file bucket in defined object storage
 func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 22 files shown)


pkg/report/handler.go
```diff
@@ -28,3 +35,45 @@ func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+
+		url, err := newPreSignedPutURL(loadTestName)
+		if nil != err {
+			render.Render(w, r, khttp.ErrResponse(http.StatusInternalServerError, err.Error()))
+			return
+		}
+
+		proxyReq, _ := http.NewRequest(r.Method, url.String(), r.Body)
```
aleksandrzhiliaev, line 64: we should use `NewRequestWithContext` and setup cancel timeout

pkg/report/handler.go
```diff
@@ -28,3 +35,45 @@ func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
...
+		proxyReq.ContentLength = r.ContentLength
+
+		proxyResp, _ := httpClient.Do(proxyReq)
```
aleksandrzhiliaev, line 67: please handle error

pkg/report/handler_test.go
```diff
@@ -45,3 +55,75 @@ func TestShowHandler(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// init dependencies for report package
+	minioClient, _ = minio.NewWithRegion("localhost:80", "access-key", "secret-access-key", false, "us-east1")
+	bucketName = "bucket-name"
+	expires = time.Second
+
+	for _, scenario := range scenarios {
```
aleksandrzhiliaev, line 104: It would be better to run each test separately with `t.Run()`, so it will be easier to check if some scenario is failed while `go run`

pkg/report/handler.go
diegomarangoni: I change to use the parent request, as I literally streaming it

pkg/report/handler.go
```diff
@@ -28,3 +36,54 @@ func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		url, err := newPreSignedPutURL(loadTestName)
+		if nil != err {
+			render.Render(w, r, khttp.ErrResponse(http.StatusInternalServerError, err.Error()))
+			return
+		}
+
+		proxyReq, err := http.NewRequestWithContext(r.Context(), r.Method, url.String(), r.Body)
+		proxyReq.ContentLength = r.ContentLength
```
aleksandrzhiliaev, line 66: `proxyReq` might be nil in case of `err!=nil`

pkg/report/handler.go
```diff
@@ -28,3 +36,54 @@ func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
...
+		proxyReq.Header = r.Header
+		if nil != err {
+			render.Render(w, r, khttp.ErrResponse(http.StatusInternalServerError, err.Error()))
+			return
+		}
```
aleksandrzhiliaev, line 71:
```suggestion
		}
defer proxyResp.Body.Close()		
```

pkg/report/handler.go
```diff
@@ -2,11 +2,19 @@ package report
...
 
+	khttp "github.com/hellofresh/kangal/pkg/core/http"
+	kk8s "github.com/hellofresh/kangal/pkg/kubernetes"
+
 	"github.com/go-chi/chi"
+	"github.com/go-chi/render"
+	"k8s.io/apimachinery/pkg/api/errors"
 )
 
+var httpClient = &http.Client{}
```
aleksandrzhiliaev, line 16: Please provide timeout for your HTTP client or handle timeouts via context like we are doing here:
https://github.com/hellofresh/kangal/blob/master/pkg/proxy/proxy.go#L265-L266

pkg/report/handler.go
```diff
@@ -28,3 +36,54 @@ func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
...
+		proxyReq.ContentLength = r.ContentLength
+		proxyReq.Header = r.Header
+		if nil != err {
+			render.Render(w, r, khttp.ErrResponse(http.StatusInternalServerError, err.Error()))
+			return
+		}
+
+		proxyResp, err := httpClient.Do(proxyReq)
+		if nil != err {
+			render.Render(w, r, khttp.ErrResponse(http.StatusInternalServerError, err.Error()))
```
aleksandrzhiliaev, line 75: We probably want to see an error message in logs too

pkg/report/handler_test.go
```diff
@@ -45,3 +57,81 @@ func TestShowHandler(t *testing.T) {
...
+		{
+			name:                   "LoadTest not found",
+			fakeResponseStatusCode: http.StatusOK,
+			getLoadTestsFn: func(action k8stesting.Action) (handled bool, ret runtime.Object, err error) {
+				return true, nil, k8serrors.NewNotFound(loadtestV1.Resource("loadtests"), "loadtest-name")
+			},
+			expectedStatusCode: http.StatusNotFound,
+		},
+		{
+			name:                   "S3 wrong credentials",
```
aleksandrzhiliaev, line 91: Is it possible to add test case with timeout request from https://github.com/hellofresh/kangal/pull/41/files#diff-68bb18c8eb3eeb398e20ae8dcc42ca2f50628d6d2fe61c83019516b6e846c97dR73 ?

pkg/report/handler.go
diegomarangoni: If reports are big we might get a timeout error

diegomarangoni: I added an 30 seconds timeout

pkg/proxy/server.go
```diff
@@ -84,6 +84,7 @@ func RunServer(ctx context.Context, cfg Config, rr Runner) error {
 		http.Redirect(w, r, url, http.StatusMovedPermanently)
 	})
 	r.Get("/load-test/{id}/report/*", report.ShowHandler())
+	r.Put("/load-test/{id}/report", report.PersistHandler(rr.KubeClient, rr.Logger))
```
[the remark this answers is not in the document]
diegomarangoni: S3 API requires a PUT, so I just keeping the same method.

aleksandrzhiliaev:
+1

PR #42, opened 2020-10-14
Configurable jmeter backend master and worker images

s4nji:
This PR:
- Makes master and worker image configurable
- Keeps the current defaults when unspecified: `hellofreshtech/kangal-jmeter-master:latest` and `hellofresh/kangal-jmeter-worker:master` when they are not specified

This allows automated testing of [`kangal-jmeter`](https://github.com/hellofresh/kangal-jmeter) images before they are released live.

### DIFF
--- a/pkg/backends/jmeter/spec.go
+++ b/pkg/backends/jmeter/spec.go
@@ -24,5 +24,5 @@ func BuildLoadTestSpec(overwrite bool, distributedPods int32, testFileStr, testD
 	if testFileStr == "" {
 		return lt, ErrRequireTestFile
 	}
-	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeJMeter, overwrite, distributedPods, testFileStr, testDataStr, envVarsStr, loadTestV1.ImageDetails{Image: masterImage, Tag: imageTag}, loadTestV1.ImageDetails{Image: workerImage, Tag: imageTag}, "", time.Duration(0)), nil
+	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeJMeter, overwrite, distributedPods, testFileStr, testDataStr, envVarsStr, loadTestV1.ImageDetails{}, loadTestV1.ImageDetails{}, "", time.Duration(0)), nil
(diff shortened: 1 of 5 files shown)


pkg/backends/jmeter/spec.go
```diff
@@ -24,5 +24,5 @@ func BuildLoadTestSpec(overwrite bool, distributedPods int32, testFileStr, testD
 	if testFileStr == "" {
 		return lt, ErrRequireTestFile
 	}
-	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeJMeter, overwrite, distributedPods, testFileStr, testDataStr, envVarsStr, loadTestV1.ImageDetails{Image: masterImage, Tag: imageTag}, loadTestV1.ImageDetails{Image: workerImage, Tag: imageTag}, "", time.Duration(0)), nil
+	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeJMeter, overwrite, distributedPods, testFileStr, testDataStr, envVarsStr, loadTestV1.ImageDetails{}, loadTestV1.ImageDetails{}, "", time.Duration(0)), nil
```
aleksandrzhiliaev, line 27: `loadTestV1.ImageDetails` should not be empty, all fields must be applied when we create a new spec.

s4nji: This will be added almost immediately in the [next](https://github.com/hellofresh/kangal/blob/1788d2dceb8de77c9325b4f43f167bd74bcf8519/pkg/backends/jmeter/jmeter.go#L73) [sync](https://github.com/hellofresh/kangal/blob/1788d2dceb8de77c9325b4f43f167bd74bcf8519/pkg/controller/loadtest.go#L316)

aleksandrzhiliaev: The proxy should give you valid LoadTest spec with all needed values, in controller previously it was a hack because setting specs didn't work properly. But right now it is not an issue anymore. In general, controller should not change resource specs - it should only change `Status` field.

PR #43, opened 2020-10-15
Remove SetDefaults method

aleksandrzhiliaev:
**Description:**
EES-4280

SetDefaults method was a previous hack because of an invalid spec of [crd.yaml](https://github.com/hellofresh/kangal/blob/master/charts/kangal/crd.yaml) 
This is not an issue anymore and Kangal Proxy is now able to prepare all specs for LoadTest so Kangal controller doesn't need to mutate resource specs anymore.

### DIFF
--- a/pkg/backends/locust/locust.go
+++ b/pkg/backends/locust/locust.go
@@ -35,29 +35,6 @@ type Locust struct {
 	podAnnotations     map[string]string
 }
 
-// SetDefaults mutates the LoadTest object to add default values to empty fields
-func (c *Locust) SetDefaults() error {
-	if c.loadTest.Status.Phase == "" {
-		c.loadTest.Status.Phase = loadTestV1.LoadTestCreating
-	}
-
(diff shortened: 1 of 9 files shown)


pkg/backends/locust/locust.go
```diff
@@ -35,29 +35,6 @@ type Locust struct {
 	podAnnotations     map[string]string
 }
 
-// SetDefaults mutates the LoadTest object to add default values to empty fields
-func (c *Locust) SetDefaults() error {
-	if c.loadTest.Status.Phase == "" {
-		c.loadTest.Status.Phase = loadTestV1.LoadTestCreating
-	}
```
diegomarangoni, line 42: This still being mutated?

diegomarangoni:
With this change, Controller will not be able to dictate image defaults anymore, and the lines below become unnecessary:
```
pkg/backends/locust/locust.go:31
pkg/backends/locust/locust.go:32
```

PR #44, opened 2020-10-16
Set MasterConfig and WorkerConfig spec via proxy

s4nji:
Currently loadtests have some empty or hardcoded spec (master/worker image details) when created by the proxy, and the controller later mutates this value. This behavior is not ideal; the spec should be immutable once the resource is created.

Furthermore, we have `MasterConfig` and `WorkerConfig` in the loadtest v1 spec, but they are not used; we should either remove this or expose it via the proxy.

This PR:
- Adds 4 environment variables (`JMETER_MASTER_IMAGE_NAME`, `JMETER_MASTER_IMAGE_TAG`, and the same for worker) to specify default images.
- Removes `SetDefault` function in controller which mutated loadtest spec.
- Updates the proxy to consume `backends.Config` environment variables and use this when creating a new loadtest.
- Adds `masterImage` and `workerImage` fields on `POST /load-test` endpoint in the proxy to allow users to utilize `MasterConfig` and `WorkerConfig` spec when creating a new loadtest.

### DIFF
--- a/pkg/backends/fake/fake.go
+++ b/pkg/backends/fake/fake.go
@@ -33,25 +33,12 @@ func New(kubeClientSet kubernetes.Interface, lt *loadTestV1.LoadTest, logger *za
 	}
 }
 
-// SetDefaults set default values for creating a Fake LoadTest pods
-func (c *Fake) SetDefaults() error {
+// CheckOrCreateResources check if Fake kubernetes resources have been create, if they have not been create them
+func (c *Fake) CheckOrCreateResources(ctx context.Context) error {
 	if c.loadTest.Status.Phase == "" {
 		c.loadTest.Status.Phase = loadTestV1.LoadTestCreating
 	}
 
-	if c.loadTest.Spec.MasterConfig.Image == "" {
-		c.loadTest.Spec.MasterConfig.Image = sleepImage
-	}
-
-	if c.loadTest.Spec.MasterConfig.Tag == "" {
-		c.loadTest.Spec.MasterConfig.Tag = imageTag
-	}
-
-	return nil
-}
-
-// CheckOrCreateResources check if Fake kubernetes resources have been create, if they have not been create them
-func (c *Fake) CheckOrCreateResources(ctx context.Context) error {
 	// Get the Namespace resource
 	namespace, err := c.kubeClient.CoreV1().Namespaces().Get(ctx, c.loadTest.Status.Namespace, metaV1.GetOptions{})
 	// The LoadTest resource may no longer exist, in which case we stop
--- a/pkg/backends/fake/fake.go
+++ b/pkg/backends/fake/fake.go
@@ -35,10 +35,6 @@ func New(kubeClientSet kubernetes.Interface, lt *loadTestV1.LoadTest, logger *za
 
 // CheckOrCreateResources check if Fake kubernetes resources have been create, if they have not been create them
 func (c *Fake) CheckOrCreateResources(ctx context.Context) error {
-	if c.loadTest.Status.Phase == "" {
-		c.loadTest.Status.Phase = loadTestV1.LoadTestCreating
-	}
-
 	// Get the Namespace resource
 	namespace, err := c.kubeClient.CoreV1().Namespaces().Get(ctx, c.loadTest.Status.Namespace, metaV1.GetOptions{})
 	// The LoadTest resource may no longer exist, in which case we stop
@@ -70,6 +66,10 @@ func (c *Fake) CheckOrUpdateStatus(ctx context.Context) error {
 	}
 	if err != nil {
 		return err
+	}
+
+	if c.loadTest.Status.Phase == "" {
+		c.loadTest.Status.Phase = loadTestV1.LoadTestCreating
 	}
 
 	if c.loadTest.Status.Phase == loadTestV1.LoadTestErrored {
--- a/pkg/backends/jmeter/jmeter.go
+++ b/pkg/backends/jmeter/jmeter.go
@@ -72,10 +72,6 @@ func New(kubeClientSet kubernetes.Interface, kangalClientSet clientSetV.Interfac
 // CheckOrCreateResources check if JMeter kubernetes resources have been create,
 // if they have not been create them
 func (c *JMeter) CheckOrCreateResources(ctx context.Context) error {
-	if c.loadTest.Status.Phase == "" {
-		c.loadTest.Status.Phase = loadTestV1.LoadTestCreating
-	}
-
 	JMeterServices, err := c.kubeClientSet.CoreV1().Services(c.loadTest.Status.Namespace).List(ctx, metaV1.ListOptions{})
 	if err != nil {
 		return err
@@ -140,6 +136,10 @@ func (c *JMeter) CheckOrUpdateStatus(ctx context.Context) error {
 			c.loadTest.Status.Phase = loadTestV1.LoadTestFinished
 			return nil
 		}
+	}
+
+	if c.loadTest.Status.Phase == "" {
+		c.loadTest.Status.Phase = loadTestV1.LoadTestCreating
 	}
 
 	if c.loadTest.Status.Phase == loadTestV1.LoadTestErrored {
--- a/pkg/backends/jmeter/jmeter.go
+++ b/pkg/backends/jmeter/jmeter.go
@@ -23,9 +23,10 @@ var (
 	MaxWaitTimeForPods = time.Minute * 10
 	//loadTestWorkerLabelSelector is the selector used for selecting jmeter worker resources
 	loadTestWorkerLabelSelector = fmt.Sprintf("%s=%s", loadTestWorkerPodLabelKey, loadTestWorkerPodLabelValue)
-	masterImage                 = "hellofreshtech/kangal-jmeter-master"
-	workerImage                 = "hellofreshtech/kangal-jmeter-worker"
-	imageTag                    = "latest"
+	defaultMasterImageName      = "hellofreshtech/kangal-jmeter-master"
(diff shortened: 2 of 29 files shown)


pkg/backends/fake/fake.go
```diff
@@ -33,25 +33,12 @@ func New(kubeClientSet kubernetes.Interface, lt *loadTestV1.LoadTest, logger *za
 	}
 }
 
-// SetDefaults set default values for creating a Fake LoadTest pods
-func (c *Fake) SetDefaults() error {
+// CheckOrCreateResources check if Fake kubernetes resources have been create, if they have not been create them
+func (c *Fake) CheckOrCreateResources(ctx context.Context) error {
```
aleksandrzhiliaev, line 37: Should this check `if c.loadTest.Status.Phase == "" {` go to `CheckOrUpdateStatus` instead?

pkg/backends/jmeter/jmeter.go
```diff
@@ -69,34 +70,13 @@ func New(kubeClientSet kubernetes.Interface, kangalClientSet clientSetV.Interfac
 	}
 }
 
-// SetDefaults set default values for creating a JMeter loadtest
-func (c *JMeter) SetDefaults() error {
+// CheckOrCreateResources check if JMeter kubernetes resources have been create,
+// if they have not been create them
+func (c *JMeter) CheckOrCreateResources(ctx context.Context) error {
```
aleksandrzhiliaev, line 75: Same as above

pkg/backends/locust/locust.go
```diff
@@ -35,31 +35,12 @@ type Locust struct {
 	podAnnotations     map[string]string
 }
 
-// SetDefaults mutates the LoadTest object to add default values to empty fields
-func (c *Locust) SetDefaults() error {
+// CheckOrCreateResources check for resources or create the needed resources for the loadtest type
+func (c *Locust) CheckOrCreateResources(ctx context.Context) error {
```
aleksandrzhiliaev, line 39: Same as above

pkg/backends/locust/spec.go
```diff
@@ -15,13 +17,21 @@ var (
...
-	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeLocust, overwrite, distributedPods, tags, testFileStr, "", envVarsStr, loadTestV1.ImageDetails{}, loadTestV1.ImageDetails{}, targetURL, duration), nil
+	// Use defaults if unspecified
+	masterImage := loadTestV1.ImageDetails{Image: defaultImage, Tag: defaultImageTag}
+	if masterImageRef != nil {
+		masterImage = loadTestV1.ImageDetails{
+			Image: masterImageRef.Name(),
+			Tag:   masterImageRef.Tag(),
+		}
+	}
+	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeLocust, overwrite, distributedPods, tags, testFileStr, "", envVarsStr, masterImage, loadTestV1.ImageDetails{}, targetURL, duration), nil
```
aleksandrzhiliaev, line 36: I think Locust uses same docker image for master-worker, @diegomarangoni can you confirm?

pkg/controller/test_helper.go
```diff
@@ -40,13 +42,26 @@ func CreateLoadtest(clientSet clientSetV.Clientset, pods int32, name, testFile,
 		}
 	}
 
+	loadtestSpec, err := backends.BuildLoadTestSpecByBackend(
+		loadTestType,
+		false,
```
aleksandrzhiliaev, line 47: This very hard to read, not sure maybe leave the old version with `ltObj.Spec.` ?

pkg/proxy/request.go
```diff
@@ -140,7 +143,19 @@ func fromHTTPRequestToLoadTestSpec(r *http.Request, logger *zap.Logger) (apisLoa
 		return apisLoadTestV1.LoadTestSpec{}, fmt.Errorf("error getting %q from request: %w", duration, err)
 	}
 
-	return backends.BuildLoadTestSpecByBackend(ltType, o, dp, tagList, tf, td, ev, turl, dur)
+	masterImageRef, err := getMasterImageRef(r)
```
aleksandrzhiliaev, line 146: This should be defined in environment variables(https://github.com/hellofresh/kangal/blob/master/pkg/backends/locust/config.go#L5-L6), and not in HTTP request.

pkg/backends/locust/spec.go
diegomarangoni: Yes, they both use same image

pkg/controller/test_helper.go
s4nji: We cannot define `ltObj.Spec.*` explicitly here; we need to call `backends.BuildLoadTestSpecByBackend()` to obtain the default master/worker config from it.

s4nji: The test was previously working because `SetDefaults()` was updating the Spec with it.  
Now that we remove `SetDefaults()`, the spec needs to have MasterConfig and WorkerConfig defined somewhere; hence the use of `backends.BuildLoadTestSpecByBackend()` here.

pkg/proxy/request.go
s4nji: Then I believe the environment variable must be removed; this config only exists in the controller side.
We established that we must not update the Spec from the controller side, so it must be configured from the proxy.

pkg/backends/jmeter/jmeter.go
s4nji: Done via 9f59c9d.

pkg/backends/fake/fake.go
s4nji: Done via 9f59c9d.

pkg/backends/locust/locust.go
s4nji: Done via 9f59c9d.

pkg/backends/locust/spec.go
s4nji: Resolved via cfc47ab.

pkg/proxy/request.go
aleksandrzhiliaev: Since the proxy is responsible for creating load test specs those env vars should be also accessible there

aleksandrzhiliaev: But I believe we should not expose it on HTTP request

diegomarangoni: For me, either the proxy should expose this or remove those properties from the Spec. Otherwise the only way to use this feature is by manually creating LoadTest resources, encouraging users to not use Proxy.

pkg/backends/jmeter/resources.go
```diff
@@ -175,7 +175,12 @@ func (c *JMeter) NewTestdataConfigMap() ([]*coreV1.ConfigMap, error) {
 func (c *JMeter) NewPod(i int, configMap *coreV1.ConfigMap, podAnnotations map[string]string) *coreV1.Pod {
 	loadtest := c.loadTest
 	optionalVolume := true
-	WorkerConfig := loadtest.Spec.WorkerConfig
+
+	imageRef := fmt.Sprintf("%s:%s", loadtest.Spec.WorkerConfig.Image, loadtest.Spec.WorkerConfig.Tag)
+	if imageRef == "" {
```
diegomarangoni, line 180: imageRef will never be empty

```suggestion
	imageRef := fmt.Sprintf("%s:%s", loadtest.Spec.WorkerConfig.Image, loadtest.Spec.WorkerConfig.Tag)
	if imageRef == ":" {
```

pkg/proxy/request.go
s4nji: > Since the proxy is responsible for creating load test specs those env vars should be also accessible there

This is now done via 2aa6dd5 829ac60; proxy will now use the same environment variables and use it when creating new loadtests.

>> But I believe we should not expose it on HTTP request

> For me, either the proxy should expose this or remove those properties from the Spec. Otherwise the only way to use this feature is by manually creating LoadTest resources, encouraging users to not use Proxy.

I agree with @diegomarangoni; I think we should either use it (i.e. expose it via the proxy) or remove this from the spec as it has no function.

pkg/backends/jmeter/resources.go
s4nji: Fixed via aec7525, thanks!

pkg/backends/locust/spec.go
```diff
@@ -15,13 +17,31 @@ var (
...
+		imageTag = config.ImageTag
+	}
+
+	// Use loadtest data received from proxy if available
+	if masterImageRef != nil {
+		imageName = masterImageRef.Name()
+		imageTag = masterImageRef.Tag()
+	}
+
+	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeLocust, overwrite, distributedPods, tags, testFileStr, "", envVarsStr, loadTestV1.ImageDetails{Image: imageName, Tag: imageTag}, loadTestV1.ImageDetails{Image: imageName, Tag: imageTag}, targetURL, duration), nil
```
diegomarangoni, line 46:
```suggestion
	imageDetails := loadTestV1.ImageDetails{
		Image: imageName,
		Tag:   imageTag,
	}

	var testDataStr string

	return loadTestV1.NewSpec(loadTestV1.LoadTestTypeLocust,
		overwrite,
		distributedPods,
		tags,
		testFileStr,
		testDataStr,
		envVarsStr,
		imageDetails,
		imageDetails,
		targetURL,
		duration), nil
```

diegomarangoni: To make it more readable :)

s4nji: Done via 24f5a58 👍

s4nji:
Closing, as this PR is split to #48 and #49.

PR #45, opened 2020-10-20
Add configmaps GET permission

diegomarangoni:
EES-4318

With the addition of Locust implementation, now Kangal requires an extra permission, a GET for ConfigMaps.


ElenaForester:
+1

ElenaForester:
+1

PR #46, opened 2020-10-21
Fix ingress annotation in Kangal chart

aleksandrzhiliaev:
We have an issue with Ingress annotations not being applied . So we found a problem in 
https://github.com/hellofresh/kangal/blob/master/charts/kangal/templates/ingress.yaml#L20

https://github.com/hellofresh/kangal/blob/master/charts/kangal/values.yaml#L27-L30

Should be:
`{{- with $value.ingress.annotations }}`

Instead of:
`{{- with $value.annotations }}`

### DIFF
--- a/charts/kangal/Chart.yaml
+++ b/charts/kangal/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - performance tests
   - tests runner
 name: kangal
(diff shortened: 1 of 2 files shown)


ElenaForester:
+1

PR #48, opened 2020-10-23
Prevent controller from changing spec

s4nji:
This PR is split from https://github.com/hellofresh/kangal/pull/44.

Currently loadtests have some empty or hardcoded spec (master/worker image details) when created by the proxy, and the controller later mutates this value. This behavior is not ideal; the spec should be immutable once the resource is created.

This PR:
- Removes `SetDefault` in the controller which mutates loadtest spec.
- Updates the proxy to consume `backends.Config` environment variables and use this when creating a new loadtest.
  - This is done so the spec will never have empty spec fields
- Adds 4 environment variables (`JMETER_MASTER_IMAGE_NAME`, `JMETER_MASTER_IMAGE_TAG`, and the same for worker) to specify default images.
  - This is done to allow users to specify custom images used for the loadtest backend.
  - It was already available in locust, this PR simply adds the same capability for JMeter.

### DIFF
--- a/charts/kangal/Chart.yaml
+++ b/charts/kangal/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - performance tests
   - tests runner
 name: kangal
-version: 1.3.0
+version: 1.4.0
 home: https://github.com/hellofresh/kangal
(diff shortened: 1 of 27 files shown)


diegomarangoni:
+1

PR #49, opened 2020-10-23
Expose loadtest master and worker config via proxy

s4nji:
This PR is split from #44.  

We have `MasterConfig` and `WorkerConfig` in the loadtest v1 spec, but they are not used; we should either remove this or expose it via the proxy.

This PR:
- Adds two additional fields (`workerImage` and `masterImage`) to `POST /load-test` endpoint in the proxy to allow users to configure `MasterConfig` and `WorkerConfig` field in the spec when creating a new loadtest.

#### Notes
Base is currently set to `patch/immutable-spec` as this PR relies on it (#48); once it's merged to master this PR will update the base branch to `master`.


diegomarangoni:
What are the use cases for having per load test images?

s4nji:
> What are the use cases for having per load test images?

The only thing I can think of is to test new images without having to rebuild or restart kangal.  
For the record, I still believe both config should be removed from the spec, but I think we need to do and plan it properly if we do that as perhaps _v2_ spec.

diegomarangoni:
> The only thing I can think of is to test new images without having to rebuild or restart kangal.

You can set a custom image once using ENV and then push to this image and restart the test.

> For the record, I still believe both config should be removed from the spec, but I think we need to do and plan it properly if we do that as perhaps v2 spec.

We can leave the PR open for now until we have a proper discussion about this feature.
Merging to discontinue later doesn’t make much sense.

PR #51, opened 2020-10-28
Change tar report rendering to use different approach

diegomarangoni:
EES-4393

Current way of untar and render report files is leading to some assets failing to download.
This PR changes the approach to locally untar reports and serve from there.


pkg/report/handler.go
```diff
@@ -28,6 +33,15 @@ func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	tmpDir := fmt.Sprintf("%s/kangal", strings.TrimRight(os.TempDir(), "/"))
+
+	go func() {
+		ticker := time.NewTicker(5 * time.Hour)
+		for range ticker.C {
+			os.RemoveAll(tmpDir)
+		}
+	}()
```
diegomarangoni, line 43: This is to ensure the local cache is always fresh. This way is not necessary a container restart to cleanup.

kirillbilchenko:
+1

PR #53, opened 2020-11-06
retrieve logs from worker pods

s-radyuk:
EES-4414

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -12,13 +12,13 @@ You can make basic troubleshooting using Kangal API endpoints or either explorin
 
 - Get status of the load test 
 ```bash
-curl --location --request GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/' 
+curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/' 
 ```
 - Get logs from the master pod
 ```bash
-curl --location --request GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs' 
+curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs' 
 ```
 - Get logs from the worker pod
 ```bash
-curl --location --request GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs/loadtest-worker-000' 
+curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs/loadtest-worker-000' 
 ```
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -9,3 +9,16 @@ To troubleshoot Kangal you will need access to your Kubernetes cluster.
 
 ## Problems with a specific load test
 You can make basic troubleshooting using Kangal API endpoints or either exploring load test Pods if is the case of your backend.
+
+- Get status of the load test 
+```bash
+curl --location --request GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/' 
+```
+- Get logs from the master pod
+```bash
+curl --location --request GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs' 
+```
+- Get logs from the worker pod
+```bash
+curl --location --request GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs/loadtest-worker-000' 
+```
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -20,5 +20,5 @@ curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs'
 ```
 - Get logs from the worker pod
 ```bash
-curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs/loadtest-worker-000' 
+curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs/0' 
 ```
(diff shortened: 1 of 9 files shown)


docs/troubleshooting.md
```diff
@@ -9,3 +9,16 @@ To troubleshoot Kangal you will need access to your Kubernetes cluster.
 
 ## Problems with a specific load test
 You can make basic troubleshooting using Kangal API endpoints or either exploring load test Pods if is the case of your backend.
+
+- Get status of the load test 
+```bash
+curl --location --request GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/'
```
diegomarangoni, line 15: is this `--location` param necessary?

s-radyuk: No, it's a copy-paste leftover, removed.

docs/troubleshooting.md
```diff
@@ -9,3 +9,16 @@ To troubleshoot Kangal you will need access to your Kubernetes cluster.
...
+```bash
+curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/' 
+```
+- Get logs from the master pod
+```bash
+curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs' 
+```
+- Get logs from the worker pod
+```bash
+curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs/loadtest-worker-000'
```
diegomarangoni, line 23:
```suggestion
curl -X GET 'http://${KANGAL_PROXY_ADDRESS}/load-test/loadtest-random-name/logs/worker-000' 
```
I think it will be better if we not repeat the `loadtest-` prefix again here, it's too much repetition.

pkg/kubernetes/client.go
```diff
@@ -149,10 +150,10 @@ func (c *Client) CountActiveLoadTests(ctx context.Context) (int, error) {
 	return counter, nil
 }
 
-// GetMasterPodLogs is making an assumptions that we only care about the logs
+// GetMasterPodRequest is making an assumptions that we only care about the logs
 // from the most recently created pod. It gets the pods associated with
 // the master job and returns the request that is used for getting the logs
-func (c *Client) GetMasterPodLogs(ctx context.Context, namespace string) (*restClient.Request, error) {
+func (c *Client) GetMasterPodRequest(ctx context.Context, namespace string) (*restClient.Request, error) {
```
diegomarangoni, line 156: what if we reuse the same function and just change the selector?

pkg/kubernetes/client.go
```diff
@@ -19,6 +19,7 @@ import (
 
 var (
 	loadTestMasterLabelSelector = "app=loadtest-master"
+	loadTestWorkerLabelSelector = "app=loadtest-worker-pod"
```
diegomarangoni, line 22: where is this being used?

pkg/kubernetes/client.go
```diff
@@ -166,6 +167,11 @@ func (c *Client) GetMasterPodLogs(ctx context.Context, namespace string) (*restC
 	return c.kubeClient.CoreV1().Pods(namespace).GetLogs(podID, &coreV1.PodLogOptions{}), nil
 }
 
+// GetWorkerPodRequest is used for getting the logs from worker pod
+func (c *Client) GetWorkerPodRequest(ctx context.Context, namespace, workerPodID string) *restClient.Request {
+	return c.kubeClient.CoreV1().Pods(namespace).GetLogs(workerPodID, &coreV1.PodLogOptions{})
```
diegomarangoni, line 172: this not going to work for all backends, Locust uses random worker suffix on name.

my suggestion is to list all workers Pods (maybe have to use labels), sort then alphabetically, and use the index order to show the users the logs.

pkg/kubernetes/client.go
s-radyuk: Now it's used for both JMeter and Locust backends to specify the worker pods.

pkg/kubernetes/client_test.go
```diff
@@ -373,10 +374,23 @@ func TestGetMasterPodLogs(t *testing.T) {
 	// to easily mock this funciton like there is for "ListPods". To do this We would
 	// need to wright our own `FakePod` package, and that doesn't seem worth it.
 	c = NewClient(loadtestClientset.KangalV1().LoadTests(), client, logger)
-	_, err = c.GetMasterPodLogs(ctx, "namespace")
+	_, err = c.GetMasterPodRequest(ctx, "namespace")
 	assert.Nil(t, err)
 }
 
+//func TestGetWorkerPodLogs(t *testing.T) {
```
vgarvardt, line 381: either remove or fix this test

pkg/backends/locust/resources.go
```diff
@@ -264,7 +265,8 @@ func newWorkerJob(
 			Template: coreV1.PodTemplateSpec{
 				ObjectMeta: metaV1.ObjectMeta{
 					Labels: map[string]string{
-						"app": name,
+						"app":            name,
+						loadTestLabelKey: loadTestWorkerLabelValue,
```
diegomarangoni, line 269: this will fail, because `loadTestLabelKey` is equal "app", but there is an existing label with the same key

pkg/backends/locust/resources.go
```diff
@@ -264,7 +265,8 @@ func newWorkerJob(
 			Template: coreV1.PodTemplateSpec{
 				ObjectMeta: metaV1.ObjectMeta{
 					Labels: map[string]string{
-						"app": name,
+						"app":            name,
```
diegomarangoni, line 268:
```suggestion
						"name":            name,
```
maybe this is what is actually wrong, just need to test to see if it works

kirillbilchenko:
@s-radyuk just a question maybe it makes sense to archive the logs from all worker pods into 1 file, instead of passing the indexes? Mean when the user will try to fetch it we will go through all worker pods fetch logs and archive them into 1 single file?

diegomarangoni:
+1

s-radyuk:
@kirillbilchenko I was thinking about compiling all logs to one stream. But from my experience, usually, all workers have almost the same logs. So most of the time users will find a reason for their problem in 1st or 2nd pod logs, no need to go through all and parse a big file/stream. 
Also appending logs to 1 stream from let's say 10 pods can be quite a heavy request. And can be overwhelming for users too.

PR #55, opened 2020-11-16
Normalise Locust ENV name

diegomarangoni:
EES-4597

Locust is using a different pattern compared to JMeter.
To keep things simple, they should use the same naming pattern.

So instead of `LOCUST_IMAGE`, should be `LOCUST_IMAGE_NAME`.
This PR introduces this change and also keep compatibility with old name.


s-radyuk:
+1

s-radyuk:
+1

PR #56, opened 2020-11-16
Use proper string to int conversion

diegomarangoni:
EES-4598

GitHub CodeQL is complaining about `int32` when using a int from `Atoi`.
This PR changes the function `Atoi` to `ParseInt`.


s-radyuk:
+1

PR #57, opened 2020-11-17
adding a phase filter for GET request

s-radyuk:
EES-273
In order to see the overview of the currently existing load test, we add a filter by phase. Now you can get all the load tests of the selected phase.
Example: this request will show all currently running load tests.
`curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?phase=running'`

This filter can be combined with filtering by tags:
`curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?phase=running&tags=tag1:value1'`

### DIFF
--- a/docs/user-flow.md
+++ b/docs/user-flow.md
@@ -102,7 +102,7 @@ You can filter by `tags`
 curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?tags=tag1:value1'
 ```
 
-You can filter by `phase`
+You can filter by `phase`, possible phases are: `creating, starting, running, finished, errored`
 
 ```bash
 curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?phase=running'
--- a/docs/user-flow.md
+++ b/docs/user-flow.md
@@ -102,6 +102,17 @@ You can filter by `tags`
 curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?tags=tag1:value1'
 ```
 
+You can filter by `phase`
+
+```bash
+curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?phase=running'
+```
+
+All together
+```bash
+curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?phase=running&tags=tag1:value1'
+```
+
 And limit your search
 
 ```bash
--- a/openapi.json
+++ b/openapi.json
@@ -33,7 +33,8 @@
 						"in": "query",
 						"description": "Filter the result by current phase.",
 						"schema": {
-							"type": "string"
+							"type": "string",
+							"enum": ["creating", "starting", "running", "finished", "errored"]
 						},
 						"example": "phase:running"
 					},
--- a/openapi.json
+++ b/openapi.json
@@ -29,6 +29,15 @@
 						"example": "department:platform,team:kangal"
 					},
 					{
+						"name": "phase",
+						"in": "query",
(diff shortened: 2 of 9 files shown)


openapi.json
```diff
@@ -28,6 +28,15 @@
 						},
 						"example": "department:platform,team:kangal"
 					},
+					{
+						"name": "phase",
+						"in": "query",
+						"description": "Filter the result by current phase.",
+						"schema": {
+							"type": "string"
```
diegomarangoni, line 36: would be better to use `enum` and list all possible phases

docs/user-flow.md
```diff
@@ -102,6 +102,17 @@ You can filter by `tags`
...
+
+```bash
+curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?phase=running'
+```
+
+All together
+```bash
+curl 'http://${KANGAL_PROXY_ADDRESS}/load-test?phase=running&tags=tag1:value1'
+```
+
```
diegomarangoni, line 115: Should we also document all the possible phases?

diegomarangoni:
+1

pkg/kubernetes/client.go
```diff
@@ -137,6 +139,18 @@ func (c *Client) ListLoadTest(ctx context.Context, opt ListOptions) (*apisLoadTe
 	return loadTests, nil
 }
 
+// ListLoadTestsByPhase returns a list of loadtests filtered by phase
+func (c *Client) ListLoadTestsByPhase(list *apisLoadTestV1.LoadTestList, phase string) *apisLoadTestV1.LoadTestList {
+	filteredList := apisLoadTestV1.LoadTestList{}
+	// CRD-s currently don't support custom field selectors, so we have to iterate via all load tests and check status phase
+	for _, loadTest := range list.Items {
+		if string(loadTest.Status.Phase) == phase {
```
vgarvardt, line 147: shall we do case-insensitive search here?

pkg/proxy/proxy_test.go
```diff
@@ -202,6 +202,42 @@ func TestProxy_List(t *testing.T) {
 			expectedContentType: "application/json; charset=utf-8",
 			expectedResponse:    "{\"error\":\"missing tag label\"}\n",
 		},
+		{
+			scenario:            "invalid phase",
+			urlParams:           "phase=foo",
+			result:              &apisLoadTestV1.LoadTestList{},
+			expectedCode:        200,
+			expectedContentType: "application/json; charset=utf-8",
+			expectedResponse:    "{\"limit\":0,\"continue\":\"\",\"remain\":null,\"items\":[]}\n",
```
vgarvardt, line 211: you can use back ticks "`" to avoid manually escaping all internal double-quotes, it will make cases clearer
```suggestion
			expectedResponse:    `{"limit":0,"continue":"","remain":null,"items":[]}\n`,
```

pkg/proxy/request.go
```diff
@@ -78,6 +78,11 @@ func fromHTTPRequestToListOptions(r *http.Request) (*kubernetes.ListOptions, err
 		opt.Limit = limit
 	}
 
+	// Build phase filter.
+	if phaseVal := params.Get("phase"); phaseVal != "" {
```
vgarvardt, line 82: Do you really need to check if the value is not empty? Default value for string field is empty string, so setting it to the param value w/out checking should not change anything.

vgarvardt:
👍

PR #59, opened 2020-11-17
EES-4591 Logger returns if it is running in debug mode

vgarvardt:
Logger is aware if debug mode is enabled, so it returns corresponding flag, so that other parts can enable some debug features based on this flag.

Not used right now, but going to be used in upcoming PRs.


diegomarangoni:
+1

PR #60, opened 2020-11-18
Prepare for new backend registry pattern migration

diegomarangoni:
EES-4624
This PR is a preparation to progress each backend to the new registry pattern.


pkg/backends/internal/registry_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"testing"
+
+	kubeFake "k8s.io/client-go/kubernetes/fake"
+
+	loadTestV1 "github.com/hellofresh/kangal/pkg/kubernetes/apis/loadtest/v1"
+	kangalFake "github.com/hellofresh/kangal/pkg/kubernetes/generated/clientset/versioned/fake"
+
+	"github.com/golang/mock/gomock"
```
vgarvardt, line 11: Is there a strong reason to use this mocking library instead of `github.com/stretchr/testify/mock` that we use in many other projects and `github.com/stretchr/testify` is already vendored?

diegomarangoni: The reasons are:
- `gomock` is part of official `golang` github organization;
- it does have a generator, which is pretty nice.

And yes, `testify` is already vendored.
I prefer `gomock` but I can change, let me know

vgarvardt: What I do not like about it is exactly generated code that is part of the codebase and anyone changing interface must not forget to regenerate mocks. But ok, maybe it is the future of mocks on go.

vgarvardt:
👍

PR #63, opened 2020-11-19
EES-4591 Added general-purpose gRPC interceptors

vgarvardt:
This is the first PR in the series of experimental gRPC with REST gateway proxy API. This PR does not change or add anything into existing functionality, all changes are standalone and not currently used. Extracted from the big chunk of code to make review process a bit easier.


pkg/proxy/interceptor/chain.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// Chain is the simple unary interceptors chain implementation
+type Chain struct {
+	interceptors []grpc.UnaryServerInterceptor
+}
+
+// NewChain builds the interceptors call chain
+func NewChain(interceptors ...grpc.UnaryServerInterceptor) *Chain {
+	return &Chain{interceptors: interceptors}
```
boekkooi-fresh, line 16: Looking at line [34](https://github.com/hellofresh/kangal/pull/63/files#diff-6047618084718f6ac165a4a27e903fb2cd6aae48e38ad368b4e41ffc9add6615R34) I'm wondering why you don't do the transformation to a final chain func here?
Aka instead if having `interceptors []grpc.UnaryServerInterceptor` just have `chain  grpc.UnaryHandler`?

vgarvardt: it uses `handler` that is only available in `Interceptor` method, it can be added as additional parameter but then it will cause more complexity so I would keep instantiation simple and do all the logic in implementation

boekkooi-fresh: Ok I'm still wondering why you don't have a UnaryServerInterceptor for the handler line [35](https://github.com/hellofresh/kangal/pull/63/files#diff-6047618084718f6ac165a4a27e903fb2cd6aae48e38ad368b4e41ffc9add6615R35) seem hacky to me. 
Instead a `UnaryServerInterceptor` that just calls the handler and never calls next would be a bit nicer to read in my opinion.

vgarvardt: Tbh I took this solution somewhere from the internet a while ago and tried to refactor to make it nicer. I failed - all the solutions I tried to implement were either worse or did not work at all. So it is like it is. If you have some time - I would really appreciate if you can make it look nicer and still working.

boekkooi-fresh: I think you can just steal the concept used in [tactician](https://github.com/thephpleague/tactician/blob/master/src/CommandBus.php) but I maybe wrong :slightly_smiling_face: 
Also where are the tests?

vgarvardt: Ok, I'm removing it completely - will use this implementation instead https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/chain.go - in any case I will use other stuff from this repo.

PR #64, opened 2020-11-19
EES-4591 gRPC/REST gateway experimental API skeleton

vgarvardt:
First PR in the upcoming series of experimental gRPC/REST gateway API.

This PR creates app and CI skeleton w/out actually implementing any logic.


Makefile
```diff
@@ -5,6 +5,8 @@ WARN_COLOR=\033[33;01m
 
 VERSION ?= "0.0.0-dev-$(shell git rev-parse --short HEAD)"
 
+GOOGLEAPIS_DIR ?= "../googleapis"
```
boekkooi-fresh, line 8: Maybe add a comment so we know where to get the content for this directory?

Makefile
```diff
@@ -34,3 +36,31 @@ apply-crd:
...
+		--proto_path $(CURDIR)/proto \
+		--go_opt=paths=source_relative \
+		--go_out=$(CURDIR)/pkg/proxy/rpc/pb \
+		--go-grpc_out=$(CURDIR)/pkg/proxy/rpc/pb \
+		--grpc-gateway_out=$(CURDIR)/pkg/proxy/rpc/pb \
+		--grpc-gateway_opt=logtostderr=true \
+		--grpc-gateway_opt=paths=source_relative \
+		--openapiv2_out=$(CURDIR)/pkg/proxy/rpc/pb \
+		--openapiv2_opt=logtostderr=true \
+		$(CURDIR)/proto/*/*/*/*.proto
```
boekkooi-fresh, line 54: Do we really need this many wildcards?

cmd/api.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+type apiCmdOpts struct {
+	kubeConfig      string
+	masterURL       string
+	maxLoadTestsRun int
+}
+
+// NewAPICmd creates a new api command
+func NewAPICmd(ctx context.Context) *cobra.Command {
```
boekkooi-fresh, line 27: Why add a new command for this and not add it to the current proxy as a flag that can be enabled?

vgarvardt: Since it is experimental it will either fail, so with the new command it would be esier to clean up, or succeed and then old proxy would be easier to remove =)

Makefile
vgarvardt: agree, added

Makefile
vgarvardt: Unfortunately yes. I think right now I can use direct path to a single file, but `/*/` works only for one level and does not expand to any levels.

boekkooi-fresh: Can't you use `/**/*.proto`? or do we actually care about the amount of levels?

cmd/api.go
```diff
@@ -0,0 +1,92 @@
...
+	opts := &apiCmdOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "api",
+		Short: "Run EXPERIMENTAL proxy for accepting API requests",
```
boekkooi-fresh, line 32:
```suggestion
		Short: "Run EXPERIMENTAL grpc proxy for accepting API requests",
```

cmd/api.go
boekkooi-fresh: Mhh not sure I agree with this but let's give it a shot.

pkg/proxy/api.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil
+	})
+
+	g.Go(func() error {
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		mux := runtime.NewServeMux()
+
+		opts := []grpc.DialOption{grpc.WithInsecure()}
```
boekkooi-fresh, line 79: `WithInsecure` should we add a todo here to make this configurable?

pkg/proxy/api.go
```diff
@@ -0,0 +1,91 @@
...
+		reflection.Register(serverAPI)
+	}
+
+	grpcAddress := fmt.Sprintf(":%d", cfg.GRPC.PortAPI)
+	restAddress := fmt.Sprintf(":%d", cfg.GRPC.PortREST)
+
+	var g errgroup.Group
+
+	g.Go(func() error {
+		tcpListener, err := net.Listen("tcp", fmt.Sprintf(":%d", cfg.GRPC.PortAPI))
```
boekkooi-fresh, line 60:
```suggestion
		tcpListener, err := net.Listen("tcp", grpcAddress)
```

pkg/proxy/api.go
```diff
@@ -0,0 +1,91 @@
...
+	})
+
+	g.Go(func() error {
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		mux := runtime.NewServeMux()
+
+		opts := []grpc.DialOption{grpc.WithInsecure()}
+		err := grpcProxyV2.RegisterLoadTestServiceHandlerFromEndpoint(ctx, mux, grpcAddress, opts)
```
boekkooi-fresh, line 80: Should this not be `restAddress`?
```suggestion
		err := grpcProxyV2.RegisterLoadTestServiceHandlerFromEndpoint(ctx, mux, restAddress, opts)
```

pkg/proxy/api.go
```diff
@@ -0,0 +1,91 @@
...
+	if rr.Debug {
+		rr.Logger.Warn("Running gRPC in debug mode with server reflection registered")
+		reflection.Register(serverAPI)
+	}
+
+	grpcAddress := fmt.Sprintf(":%d", cfg.GRPC.PortAPI)
+	restAddress := fmt.Sprintf(":%d", cfg.GRPC.PortREST)
+
+	var g errgroup.Group
+
```
boekkooi-fresh, line 58: You probably want this context aware see https://godoc.org/golang.org/x/sync/errgroup#WithContext

pkg/proxy/api.go
```diff
@@ -0,0 +1,91 @@
...
+		err := grpcProxyV2.RegisterLoadTestServiceHandlerFromEndpoint(ctx, mux, grpcAddress, opts)
+		if err != nil {
+			return fmt.Errorf("could not register service Ping: %w", err)
+		}
+
+		rr.Logger.Info("Running gRPC REST server...", zap.String("addr", restAddress))
+		return http.ListenAndServe(restAddress, mux)
+	})
+
+	err := g.Wait()
```
boekkooi-fresh, line 89: Are you sure you want `Wait blocks until all function calls from the Go method have returned, then returns the first non-nil error (if any) from them.`
Should this not fail when 1 fails?

See https://godoc.org/golang.org/x/sync/errgroup#Group.Wait

Makefile
vgarvardt: nope

```
❯ make protoc
==> Compiling ProtoBuf
Could not make proto path relative: /Users/vladimir.garvardt/Projects/kangal/proto/**/*.proto: No such file or directory
make: *** [protoc] Error 1
```

pkg/proxy/api.go
vgarvardt: probably yes, but not very soon - current proxy API has no app-level protection at all

will add TODO anyway

pkg/proxy/api.go
vgarvardt: this is correct - http server listens to `restAddress`, and this is generated proxy that receives REST requests using `mux`, converts them and forwards to `grpcAddress`

pkg/proxy/api.go
vgarvardt: you're correct, will fix this

Makefile
boekkooi-fresh: Maybe set the [shell](https://github.com/hellofresh/vault-secrets-controller/blob/master/Makefile#L177)?

pkg/proxy/api.go
vgarvardt: done

pkg/proxy/api.go
vgarvardt: fixed, at least should be fixed

Makefile
vgarvardt: Nope, does not work as well and there are a lot of different options in in internet, using `find`, `for`, etc. looks like this path is expanded on the `protoc` app level that is C++ app

boekkooi-fresh:
:+1: RunAPIServer could do with some method being extracted but for an experiment it's good enough I think.

PR #65, opened 2020-11-20
JMeter implementation of registry pattern

diegomarangoni:
EES-4627
This PR implement new registry pattern into JMeter backend


s4nji:
👍

pkg/backends/jmeter/backend.go
```diff
@@ -0,0 +1,280 @@
...
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	coreListersV1 "k8s.io/client-go/listers/core/v1"
+
+	"github.com/hellofresh/kangal/pkg/backends/internal"
+	"github.com/hellofresh/kangal/pkg/core/helper"
+	loadTestV1 "github.com/hellofresh/kangal/pkg/kubernetes/apis/loadtest/v1"
+	clientSetV "github.com/hellofresh/kangal/pkg/kubernetes/generated/clientset/versioned"
+
+	"go.uber.org/zap"
```
boekkooi-fresh, line 16: Did goimport formatting go bad again?

pkg/backends/jmeter/backend.go
```diff
@@ -0,0 +1,280 @@
...
+						zap.String("pod", pod.Name),
+						zap.String("namespace", namespace.GetName()),
+					)
+					loadTestStatus.Phase = loadTestV1.LoadTestErrored
+					return nil
+				}
+			}
+
+			return nil
+		}
```
boekkooi-fresh, line 266: Can we extract this into a `func`?

pkg/backends/jmeter/backend.go
```diff
@@ -0,0 +1,280 @@
...
+func (b *Backend) SyncStatus(ctx context.Context, loadTest loadTestV1.LoadTest, loadTestStatus *loadTestV1.LoadTestStatus) error {
+	// Get the Namespace resource
+	namespace, err := b.namespaceLister.Get(loadTestStatus.Namespace)
+	if err != nil {
+		// The LoadTest resource may no longer exist, in which case we stop
+		// processing.
+		if kerrors.IsNotFound(err) {
+			loadTestStatus.Phase = loadTestV1.LoadTestFinished
+			return nil
+		}
```
boekkooi-fresh, line 215: Are we missing an err return?
```suggestion
		}
		return err
```

pkg/backends/jmeter/jmeter.go
```diff
@@ -53,223 +45,54 @@ func New(
 	logger *zap.Logger,
 	namespacesLister coreListersV1.NamespaceLister,
 	reportURL string,
-	podAnnotations, namespaceAnnotations map[string]string,
+	podAnnotations, _ map[string]string,
```
boekkooi-fresh, line 48: Why is `namespaceAnnotations` no longer used?

pkg/backends/jmeter/jmeter.go
```diff
@@ -53,223 +45,54 @@ func New(
...
-	// the Job status
-	c.loadTest.Status.Phase = getLoadTestPhaseFromJob(job.Status)
-	c.loadTest.Status.JobStatus = job.Status
-
-	return nil
+	return c.backend.SyncStatus(ctx, *c.loadTest, &c.loadTest.Status)
 }
 
-func (c *JMeter) createPodsWithTestdata(ctx context.Context, configMaps []*coreV1.ConfigMap, loadtest *loadTestV1.LoadTest, namespace string) error {
+func (b *Backend) createPodsWithTestdata(ctx context.Context, configMaps []*coreV1.ConfigMap, loadTest *loadTestV1.LoadTest, namespace string) error {
```
boekkooi-fresh, line 81: I think you forgot to move this to the right file.

pkg/backends/jmeter/resources.go
```diff
@@ -56,7 +54,7 @@ var (
 )
 
 // NewJMeterSettingsConfigMap creates a new configmap which holds jmeter config
-func (c *JMeter) NewJMeterSettingsConfigMap() *coreV1.ConfigMap {
+func (b *Backend) NewJMeterSettingsConfigMap() *coreV1.ConfigMap {
```
boekkooi-fresh, line 57: Why was this changed? This breaks BC at the moment right? 
(Same question for other `func (c *JMeter)` to `func (b *Backend) ` changes.)

pkg/backends/jmeter/jmeter.go
diegomarangoni: Namespace are created by controller itself, so backends don't need it.
This will be removed on the future

pkg/backends/jmeter/resources.go
diegomarangoni: `JMeter` now only have aliases to `Backend` functions, so this needs to be moved to new `Backend` struct as well.
Once the switch to this new registry pattern is done, all this `JMeter` struct can be deleted.

boekkooi-fresh: I know but right now the tests no longer cover `JMeter` meaning people can break tests easily for `JMeter` but keep `Backend` tests passing. 
This in turn leads to other issues. I would advice to move the tests once you remove `JMeter` and we switch to using `Backend` directly

diegomarangoni: That's the plan, but I also had to adapt the test, I will tag you on the related code.

pkg/backends/jmeter/resources_test.go
```diff
@@ -93,19 +93,20 @@ func TestGetNamespaceFromInvalidName(t *testing.T) {
...
+				Tag:   defaultMasterImageTag,
+			},
+			WorkerConfig: loadtestv1.ImageDetails{
+				Image: defaultWorkerImageName,
+				Tag:   defaultWorkerImageTag,
 			},
 		},
+	}
+
+	c := &Backend{
```
diegomarangoni, line 109: @boekkooi-fresh note that here I had to change from `JMeter` to `Backend` instead, to actually test those implementations.

pkg/backends/jmeter/backend.go
```diff
@@ -0,0 +1,306 @@
...
+	if err != nil {
+		return err
+	}
+
+	if len(pods.Items) != int(*loadTest.Spec.DistributedPods) {
+		return nil
+	}
+
+	for _, pod := range pods.Items {
+		if pod.Status.Phase != "Running" {
```
aleksandrzhiliaev, line 240: Should those pod status phases be consts?

pkg/backends/jmeter/backend.go
```diff
@@ -0,0 +1,306 @@
...
+				loadTestStatus.Phase = loadTestV1.LoadTestFinished
+				return nil
+			}
+			loadTestStatus.Phase = getLoadTestStatusPhaseByPod(pod)
+			if loadTestV1.LoadTestErrored == loadTestStatus.Phase {
+				b.logger.Info(
+					"One of containers is unhealthy, marking LoadTest as errored",
+					zap.String("loadTest", loadTest.GetName()),
+					zap.String("pod", pod.Name),
+					zap.String("namespace", namespace.GetName()),
```
aleksandrzhiliaev, line 253: Here `namespace` and load test name are identical.

pkg/backends/jmeter/jmeter.go
```diff
@@ -53,243 +41,35 @@ func New(
...
-		}
-		c.logger.Info(
-			"Created JMeter resources",
-			zap.String("LoadTest", c.loadTest.GetName()),
-			zap.String("namespace", c.loadTest.Status.Namespace),
-		)
-
-	}
-	return nil
+	return c.backend.Sync(ctx, *c.loadTest, c.reportURL)
```
aleksandrzhiliaev, line 68: I would also add debug logging on each `CheckOrCreateResources` or we can move `debug` log to `Sync` method.

pkg/backends/jmeter/jmeter.go
```diff
@@ -53,243 +41,35 @@ func New(
...
-	if status.Active > 0 {
-		return loadTestV1.LoadTestRunning
-	}
-
-	if status.Succeeded == 0 && status.Failed == 0 {
-		return loadTestV1.LoadTestStarting
-	}
-
-	return loadTestV1.LoadTestFinished
+	return c.backend.SyncStatus(ctx, *c.loadTest, &c.loadTest.Status)
```
aleksandrzhiliaev, line 74: I would also add debug logging on each `CheckOrUpdateStatus` or we can move debug log to `SyncStatus` method.

aleksandrzhiliaev (commented):
Left small suggestions. Do we cover that JMeter backend by tests?

diegomarangoni:
> Left small suggestions. Do we cover that JMeter backend by tests?

Yes, we do

pkg/backends/jmeter/jmeter.go
diegomarangoni: which log are you referring to? I didn’t remove any log, or did?

pkg/backends/jmeter/jmeter.go
diegomarangoni: which log are you referring to? I didn’t remove any log, or did?

pkg/backends/jmeter/jmeter.go
aleksandrzhiliaev: I think we didn't have any. I thought maybe as addition, might be useful when you debugging controller.

boekkooi-fresh: Sounds like a good idea for another Ticket + PR

aleksandrzhiliaev:
+1

PR #66, opened 2020-11-24
EES-4641 Use k8s errors to find out about LT not found

vgarvardt:
Use k8s errors to find out about LT not found - use built-in k8s errors to find out if object is not found instead of type casting errors and checking fields manually

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -8,14 +8,13 @@ import (
 	"io"
 	"net/http"
 
-	"github.com/hellofresh/kangal/pkg/backends"
-
 	"github.com/go-chi/chi"
 	"github.com/go-chi/render"
 	"go.uber.org/zap"
 	k8sAPIErrors "k8s.io/apimachinery/pkg/api/errors"
 	restClient "k8s.io/client-go/rest"
 
+	"github.com/hellofresh/kangal/pkg/backends"
 	loadtest "github.com/hellofresh/kangal/pkg/controller"
 	cHttp "github.com/hellofresh/kangal/pkg/core/http"
 	mPkg "github.com/hellofresh/kangal/pkg/core/middleware"
@@ -162,7 +161,7 @@ func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
 				// Remove the old tests
 				err := p.kubeClient.DeleteLoadTest(ctx, item.Name)
 				if err != nil {
-					logger.Error("Could not delete load test with error:", zap.Error(err))
+					logger.Error("Could not delete load test with error", zap.Error(err))
 					render.Render(w, r, cHttp.ErrResponse(http.StatusConflict, err.Error()))
 					return
 				}
@@ -220,7 +219,7 @@ func (p *Proxy) Delete(w http.ResponseWriter, r *http.Request) {
 
 	err := p.kubeClient.DeleteLoadTest(ctx, ltID)
 	if err != nil {
-		logger.Error("Could not delete load test with error:", zap.Error(err))
+		logger.Error("Could not delete load test with error", zap.Error(err))
 		render.Render(w, r, cHttp.ErrResponse(http.StatusBadRequest, err.Error()))
 		return
 	}
@@ -242,7 +241,7 @@ func (p *Proxy) Get(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		logger.Error("Could not get load test info with error:", zap.Error(err))
 
-		if k8sErr, ok := err.(k8sAPIErrors.APIStatus); ok && k8sErr.Status().Code == http.StatusNotFound {
+		if k8sAPIErrors.IsNotFound(err) {
 			render.Render(w, r, cHttp.ErrResponse(http.StatusNotFound, err.Error()))
 			return
 		}
@@ -275,7 +274,7 @@ func (p *Proxy) GetLogs(w http.ResponseWriter, r *http.Request) {
 
 	loadTest, err := p.kubeClient.GetLoadTest(ctx, ltID)
 	if err != nil {
-		logger.Error("Could not get load test info with error:", zap.Error(err))
+		logger.Error("Could not get load test info with error", zap.Error(err))
 		render.Render(w, r, cHttp.ErrResponse(http.StatusBadRequest, err.Error()))
 		return
 	}
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -239,7 +239,7 @@ func (p *Proxy) Get(w http.ResponseWriter, r *http.Request) {
 
 	result, err := p.kubeClient.GetLoadTest(ctx, ltID)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -37,11 +37,7 @@ jobs:
         env:
           GOOGLEAPIS_DIR: "./googleapis"
         run: |
-          go install \
-              github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
-              github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 \
-              google.golang.org/protobuf/cmd/protoc-gen-go \
-              google.golang.org/grpc/cmd/protoc-gen-go-grpc
+          make protoc-plugins
```
boekkooi-fresh, line 40: Why not make this a requirement for `protoc` in the makefile?

docs/README.md
```diff
@@ -93,22 +93,68 @@ kubectl apply -f charts/kagal/crd.yaml
...
+```bash
+make protoc-plugins
+```
+
+#### 3.3 Checkout required protobuf definitions
+
+In order to allow 3rd party API definitions you need to have them locally:
+
+```bash
+# go outside of the kangas source path
```
s-radyuk, line 125:
```suggestion
# go outside of the kangal source path
```

Makefile
```diff
@@ -41,6 +41,16 @@ apply-crd:
...
 	@kubectl apply -f charts/kangal/crd.yaml
 
+# Install required protoc plugins
+protoc-plugins:
+	@printf "$(OK_COLOR)==> Installing required protoc plugins$(NO_COLOR)\n"
+	@go install \
+	   github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
+	   github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 \
+	   google.golang.org/protobuf/cmd/protoc-gen-go \
+	   google.golang.org/grpc/cmd/protoc-gen-go-grpc
```
boekkooi-fresh, line 51: Do we really need to install things system wide? Can't we use something like the following to also ensure the correct version.
```
	BIN_DIR=...
	BUILD_DIR=$$(mktemp -d); \
	cd "$${BUILD_DIR}"; \
	go mod init protoc-plugins; \
	go get "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway@v2.0.1"; \
	go build -mod=readonly -o "$${BIN_DIR}/protoc-gen-grpc-gateway" github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway; \
	rm -rf "$${BUILD_DIR}"
	"$${BIN_DIR}/protoc-gen-grpc-gateway" --version
```

pkg/proxy/proxy.go
```diff
@@ -242,7 +241,7 @@ func (p *Proxy) Get(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		logger.Error("Could not get load test info with error:", zap.Error(err))
 
-		if k8sErr, ok := err.(k8sAPIErrors.APIStatus); ok && k8sErr.Status().Code == http.StatusNotFound {
+		if k8sAPIErrors.IsNotFound(err) {
```
boekkooi-fresh, line 244: Seems like this is unrelated to the protoc makefile changes can you split this PR please.

.github/workflows/release.yml
vgarvardt: `make protoc` should be called for every proto file change, calling `go install ...` every time would take much more time

pkg/proxy/proxy.go
vgarvardt: It is already in another dedicated commit.

Makefile
vgarvardt: These binaries are `protoc` plugins, installing them like this makes the whole set up much more complicated. And we expect `protoc` installed globally as well.

boekkooi-fresh: Should we have protoc installed globally? Are we going to support protoc v1, v2, v3 etc?
By not providing versions we actually get a build system that will very every single time something is build.

.github/workflows/release.yml
boekkooi-fresh: Then use make files properly and only build what is needed. For example:
```
# Protoc is used to generate code from a protobuf file
PROTOC_VERSION=3.11.4
PROTOC_ZIP=protoc-$(PROTOC_VERSION)-linux-x86_64.zip
ifeq ($(GOHOSTOS)_$(GOHOSTARCH),darwin_amd64)
    PROTOC_ZIP=protoc-$(PROTOC_VERSION)-osx-x86_64.zip
endif
PROTOC_RELEASES_URI=https://github.com/protocolbuffers/protobuf/releases/download
PROTOC_DOWNLOAD=$(PROTOC_RELEASES_URI)/v$(PROTOC_VERSION)/$(PROTOC_ZIP)

tools/protoc = $(BIN_DIR)/protoc
$(tools/protoc):
	mkdir -p "$(@D)"
	curl --fail --output "$@.zip" -L "$(PROTOC_DOWNLOAD)"
	go run tools/unzip.go "$@.zip" "$@" "bin/protoc"
	rm -f "$@.zip"
```

Makefile
vgarvardt: We are using latest stable `protoc` version that is `3.x` - it is pinned in the CI and in the dev instruction.

Makefile
```diff
@@ -9,7 +9,7 @@ VERSION ?= "0.0.0-dev-$(shell git rev-parse --short HEAD)"
 # since we're using github.com/grpc-ecosystem/grpc-gateway it needs to be manually cloned and put somewhere
 # close to the project and this env var should point to "third_party/googleapis" subdir.
 # IMPORTANT - check currently used version of github.com/grpc-ecosystem/grpc-gateway to use exactly the same tag
-GOOGLEAPIS_DIR ?= "../googleapis"
+GOOGLEAPIS_DIR ?= "../grpc-gateway/third_party/googleapis"
```
diegomarangoni, line 12: what do you think if we use submodules to clone third parties?
this way we ensure `googleapis` is at correct version

vgarvardt: git submodules is the worst thing I've ever used for vendoring deps, I'll think how to do this in an easier way

vgarvardt:
Ok, `protoc` stuff turned to be more than just updating README, so it goes into https://github.com/hellofresh/kangal/pull/67 to unblock small real code change as every merge to master requires rebase.

pkg/proxy/proxy.go
```diff
@@ -242,7 +241,7 @@ func (p *Proxy) Get(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		logger.Error("Could not get load test info with error:", zap.Error(err))
```
diegomarangoni, line 242:
```suggestion
		logger.Error("Could not get load test info with error", zap.Error(err))
```
one more here

aleksandrzhiliaev:
👍

PR #68, opened 2020-11-24
EES-4641 Implemented gRPC Get method to get load tests info by name

vgarvardt:
The first and the simplest API method implemented. Includes unit and integration tests using REST gateway.

### DIFF
--- a/pkg/proxy/proxy_integration_test.go
+++ b/pkg/proxy/proxy_integration_test.go
@@ -449,6 +449,7 @@ func TestIntegrationGetLoadtest(t *testing.T) {
 	})
 
 	t.Run("Ensure loadtest gRPC/REST gateway GET response is correct", func(t *testing.T) {
+		require.NotEmpty(t, restBody)
 		t.Logf("gRPC/REST gateway response: %s", restBody)
 
 		dat := new(grpcProxyV2.GetResponse)
--- a/pkg/proxy/proxy_integration_test.go
+++ b/pkg/proxy/proxy_integration_test.go
@@ -13,12 +13,13 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	coreV1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8sAPIErrors "k8s.io/apimachinery/pkg/api/errors"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	testhelper "github.com/hellofresh/kangal/pkg/controller"
 	apisLoadTestV1 "github.com/hellofresh/kangal/pkg/kubernetes/apis/loadtest/v1"
 	clientSetV "github.com/hellofresh/kangal/pkg/kubernetes/generated/clientset/versioned"
+	grpcProxyV2 "github.com/hellofresh/kangal/pkg/proxy/rpc/pb/grpc/proxy/v2"
 )
 
 var (
@@ -342,7 +343,7 @@ func TestIntegrationDeleteLoadtest(t *testing.T) {
 		// It means http.StatusNotFound is a good result for Cleanup
 		// If Cleanup returns some other error we should assert it
 		err := testhelper.DeleteLoadTest(clientSet, expectedLoadtestName, t.Name())
-		statusErr, ok := err.(*errors.StatusError)
+		statusErr, ok := err.(*k8sAPIErrors.StatusError)
 		if !ok || statusErr.ErrStatus.Code != http.StatusNotFound {
 			assert.NoError(t, err)
 			return
@@ -391,7 +392,10 @@ func TestIntegrationGetLoadtest(t *testing.T) {
 		require.NoError(t, err)
 	})
 
-	var respBody []byte
+	var (
+		httpBody []byte
+		restBody []byte
+	)
 
 	t.Run("Get loadtest details", func(t *testing.T) {
 		req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://localhost:%d/load-test/%s", httpPort, expectedLoadtestName), nil)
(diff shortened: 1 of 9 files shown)


pkg/proxy/proxy_integration_test.go
```diff
@@ -415,6 +430,41 @@ func TestIntegrationGetLoadtest(t *testing.T) {
...
 		assert.Equal(t, false, dat.HasTestData)
 	})
+
+	t.Run("Get loadtest details from gRPC/REST gateway", func(t *testing.T) {
+		req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://localhost:%d/v2/load-test/%s", restPort, expectedLoadtestName), nil)
+		require.NoError(t, err, "Could not create GET request")
+
+		res, err := http.DefaultClient.Do(req)
+		require.NoError(t, err, "Could not send GET request")
+		require.Equal(t, http.StatusOK, res.StatusCode)
```
boekkooi-fresh, line 440: Should this not be placed after the `defer` on line 442? or changed to be an `assert` call

pkg/proxy/proxy_integration_test.go
```diff
@@ -415,6 +430,41 @@ func TestIntegrationGetLoadtest(t *testing.T) {
...
+
+		defer func() {
+			err := res.Body.Close()
+			assert.NoError(t, err)
+		}()
+
+		restBody, err = ioutil.ReadAll(res.Body)
```
boekkooi-fresh, line 447: Is there an assertion regarding `restBody` missing?

pkg/proxy/service.go
```diff
@@ -3,24 +3,106 @@ package proxy
...
+	case apisLoadTestV1.LoadTestFinished:
+		return grpcProxyV2.LoadTestPhase_LOAD_TEST_PHASE_FINISHED
+	case apisLoadTestV1.LoadTestErrored:
+		return grpcProxyV2.LoadTestPhase_LOAD_TEST_PHASE_ERRORED
+	}
+
+	return grpcProxyV2.LoadTestPhase_LOAD_TEST_PHASE_UNSPECIFIED
+}
+
+func (s *implLoadTestServiceServer) typeToGRPC(t apisLoadTestV1.LoadTestType) grpcProxyV2.LoadTestType {
```
boekkooi-fresh, line 84: Maybe make this a package level func?
```suggestion
func typeToGRPC(t apisLoadTestV1.LoadTestType) grpcProxyV2.LoadTestType {
```

pkg/proxy/service.go
```diff
@@ -3,24 +3,106 @@ package proxy
...
 
 // List searches and returns load tests by given filters
 func (s *implLoadTestServiceServer) List(context.Context, *grpcProxyV2.ListRequest) (*grpcProxyV2.ListResponse, error) {
 	return new(grpcProxyV2.ListResponse), nil
 }
 
-// Get returns load test by given name
-func (s *implLoadTestServiceServer) Get(context.Context, *grpcProxyV2.GetRequest) (*grpcProxyV2.GetResponse, error) {
-	return new(grpcProxyV2.GetResponse), nil
+func (s *implLoadTestServiceServer) phaseToGRPC(p apisLoadTestV1.LoadTestPhase) grpcProxyV2.LoadTestPhase {
```
boekkooi-fresh, line 67: Maybe make this a package level func?
```suggestion
func phaseToGRPC(p apisLoadTestV1.LoadTestPhase) grpcProxyV2.LoadTestPhase {
```

pkg/proxy/service.go
```diff
@@ -3,24 +3,106 @@ package proxy
...
+	case apisLoadTestV1.LoadTestTypeFake:
+		return grpcProxyV2.LoadTestType_LOAD_TEST_TYPE_FAKE
+	case apisLoadTestV1.LoadTestTypeLocust:
+		return grpcProxyV2.LoadTestType_LOAD_TEST_TYPE_LOCUST
+	}
+
+	return grpcProxyV2.LoadTestType_LOAD_TEST_TYPE_UNSPECIFIED
+}
+
+func (s *implLoadTestServiceServer) tagsToGRPC(tt apisLoadTestV1.LoadTestTags) []*grpcProxyV2.Tag {
```
boekkooi-fresh, line 97: Maybe make this a package level func?
```suggestion
func tagsToGRPC(tt apisLoadTestV1.LoadTestTags) []*grpcProxyV2.Tag {
```

pkg/proxy/service_test.go
```diff
@@ -0,0 +1,99 @@
...
+				return true, &tt.loadTest, tt.ltErr
+			})
+			c := kube.NewClient(loadtestClientSet.KangalV1().LoadTests(), kubeClientSet, logger)
+
+			svc := NewLoadTestServiceServer(c)
+
+			out, err := svc.Get(ctx, &grpcProxyV2.GetRequest{Name: "aaa"})
+			assert.Equal(t, tt.out, out)
+
+			if tt.outErr != nil {
```
boekkooi-fresh, line 93: More me this type of if statement is a code smell. Who not split "error" and "success" tests from each other?

pkg/proxy/service.go
vgarvardt: fixed

pkg/proxy/service.go
vgarvardt: fixed

pkg/proxy/service.go
vgarvardt: fixed

pkg/proxy/proxy_integration_test.go
vgarvardt: It can not be missing, it can only be empty and the next sub-test asserts if it matches expectations. But anyway I added extra validation to fail early if it is empty

pkg/proxy/proxy_integration_test.go
vgarvardt: true, fixed

pkg/proxy/service_test.go
vgarvardt: changed errors asserting a bit, so now there is no `if` anymore

pkg/proxy/proxy_integration_test.go
boekkooi-fresh: Sorry but where is the assertion for `restBody`? I can't find it.

vgarvardt: in the next subtest, that asserts contents of the body, tries to decode it, etc.

vgarvardt: And here is commit that adds it https://github.com/hellofresh/kangal/pull/68/commits/412e07252c525933718bdc5ab598bc21ef0d22b8

boekkooi-fresh: Hold on ... you are share context between sub tests on the same level? Why have them as subsets at all then? This seems wrong to me.

vgarvardt: this is how it is implemented here, and this is how v1 API is being tested, I'm trying not to touch anything yet

boekkooi-fresh: Let's create a ticket to clean this up later across the tests What do you think?

vgarvardt: Sounds good to me. Created one.

boekkooi-fresh:
:+1: Thanks for the work let's hope this experiment brings us something :smile:

PR #69, opened 2020-11-25
Fixed CI release flow

vgarvardt:
Hotfix for failing build - https://github.com/hellofresh/kangal/runs/1452857789?check_suite_focus=true - clean up source tree after code was generated from protobuf

I'm working on more reliable compile-time tooling solution, this hotfix only to unblock releases.


diegomarangoni:
+1

PR #71, opened 2020-11-25
Update Helm chart-testing pipeline

diegomarangoni:
Previous image no longer exists and pipeline is failing with error below.
```
Run helm/chart-testing-action@v1.0.0
Running ct container...
Unable to find image 'quay.io/helmpack/chart-testing:v2.4.1' locally
docker: Error response from daemon: received unexpected HTTP status: 500 Internal Server Error.
See 'docker run --help'.
```
This PR updates pipeline to use latest version of `helm/chart-testing-action`.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -85,12 +85,21 @@ jobs:
       - name: Fetch history for helm linter
         run: git fetch --prune --unshallow
 
-      - name: Lint Helm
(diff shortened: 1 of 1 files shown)


vgarvardt:
It is not an image issue, but quay.io is being partially down. But the change is fine in any case IMO.

vgarvardt:
👍

PR #72, opened 2020-11-26
Implement Locust registry pattern

diegomarangoni:
EES-4626
As part of registry pattern implementation, this PR implements into Locust backend.


pkg/backends/locust/backend.go
```diff
@@ -0,0 +1,240 @@
+package locust
+
+import (
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+	coreV1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
```
vgarvardt, line 9: pls do like

```suggestion
	k8sAPIErrors "k8s.io/apimachinery/pkg/api/errors"
```

to avoid mixing stdlib `errors` and k8s errors

vgarvardt:
👍

PR #73, opened 2020-11-26
EES-4643 Set go_package explicitly and do not use path=source_relative

vgarvardt:
Minor improvements for protobuf tooling:
- finally found out how to use `go_package` correctly to get rid of warnings from transpiler
- use map for tags as they are map actually internally


boekkooi-fresh:
:+1:

PR #75, opened 2020-11-27
Switch to the new registry

diegomarangoni:
EES-4628
This PR switch to use the new registry pattern.
A code cleanup will come next.


pkg/backends/all/all.go
```diff
@@ -0,0 +1,10 @@
+package all
```
boekkooi-fresh, line 1: Why no move this imports to the main.go?

pkg/controller/loadtest.go
```diff
@@ -306,9 +310,9 @@ func (c *Controller) syncHandler(key string) error {
 		reportURL = fmt.Sprintf("%s/load-test/%s/report", c.cfg.KangalProxyURL, loadTest.GetName())
 	}
 
-	backend, err := backends.NewLoadTest(loadTest, c.kubeClientSet, c.kangalClientSet, c.logger, c.namespacesLister, reportURL, c.cfg.PodAnnotations, c.cfg.NamespaceAnnotations, c.cfg.Backends)
+	resolved, err := c.registry.Resolve(loadTest.Spec.Type)
```
boekkooi-fresh, line 313: For some reason the `registry` or `Resolve` name here seems off to me.
Based on [Resolve](https://github.com/hellofresh/kangal/blob/ddb64691658d562862ebb6a7e29d2defd5801718/pkg/backends/internal/registry.go#L61) I think `Get` or `GetBackend` maybe a better name here.

pkg/proxy/proxy.go
```diff
@@ -31,21 +31,19 @@ var (
...
-	kubeClient          *kube.Client
-	httpToSpecConverter loadTestSpecCreator
+	maxLoadTestsRun int
+	registry        *backends.Registry
+	kubeClient      *kube.Client
 }
 
 // NewProxy returns new Proxy handlers
-func NewProxy(cfg Config, kubeClient *kube.Client, specCreator loadTestSpecCreator) *Proxy {
+func NewProxy(maxLoadTestsRun int, registry *backends.Registry, kubeClient *kube.Client) *Proxy {
```
boekkooi-fresh, line 42: Why are we expecting a `struct` and not an interface? (same question for the other places where `*backends.Registry` is used) To me this seems to make the code a lot harder to test properly. Maybe use and interface?

pkg/backends/all/all.go
diegomarangoni: Because then I would have to repeat all imports on another packages, see this commit: https://github.com/hellofresh/kangal/pull/75/commits/fa91f60b1caf5f014e18a7d88062c5bfdba37724

boekkooi-fresh: This all package feel like a hack to me since it break the concept of the folders within backends at the moment. We have no backend/system called `all` that we support. 
Also those `packages` you link to are tests which I would argue should not break if another backend is added to the system which in this case may happen.

I possible solution would be to use an `internal` package to register all backends or to split the backends package into backends and backend.
Where backend would contains the Interfaces and backends the import for all backends in it's subdirs.

diegomarangoni: what do you think now?

boekkooi-fresh: I'm wondering if naming the package after the pattern `registry` is a good idea. What happens if you want to add another registry to the project? 
Also when I look at `registry.New` for example is unclear to me what the registry is for. Maybe it's a generic registry?

diegomarangoni: I can improve the `registry.New` to be more clear or I can move both, `registry` and `backends` into a `backend` package, but to be honest I don't like, I prefer the first option

pkg/proxy/service.go
```diff
@@ -61,53 +61,35 @@ func (s *implLoadTestServiceServer) Get(ctx context.Context, in *grpcProxyV2.Get
...
+		return grpcProxyV2.LoadTestPhase_LOAD_TEST_PHASE_ERRORED
 	}
 
 	return grpcProxyV2.LoadTestPhase_LOAD_TEST_PHASE_UNSPECIFIED
 }
 
 func typeToGRPC(t apisLoadTestV1.LoadTestType) grpcProxyV2.LoadTestType {
-	if grpcVal, found := typeToGRPCMap[t]; found {
-		return grpcVal
+	switch t {
```
boekkooi-fresh, line 85: Looks like a merge conflict

pkg/proxy/service.go
```diff
@@ -61,53 +61,35 @@ func (s *implLoadTestServiceServer) Get(ctx context.Context, in *grpcProxyV2.Get
...
 // List searches and returns load tests by given filters
 func (s *implLoadTestServiceServer) List(context.Context, *grpcProxyV2.ListRequest) (*grpcProxyV2.ListResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	return new(grpcProxyV2.ListResponse), nil
 }
 
 func phaseToGRPC(p apisLoadTestV1.LoadTestPhase) grpcProxyV2.LoadTestPhase {
-	if grpcVal, found := phaseToGRPCMap[p]; found {
-		return grpcVal
+	switch p {
```
boekkooi-fresh, line 68: Looks like a merge conflict

pkg/proxy/proxy.go
```diff
@@ -181,12 +177,24 @@ func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
...
-		logger.Warn("number of active load tests reached limit", zap.Int("current", activeLoadTests), zap.Int("limit", p.config.MaxLoadTestsRun))
+	if activeLoadTests >= p.maxLoadTestsRun {
+		logger.Warn("number of active load tests reached limit", zap.Int("current", activeLoadTests), zap.Int("limit", p.maxLoadTestsRun))
 		render.Render(w, r, cHttp.ErrResponse(http.StatusTooManyRequests, "Number of active load tests reached limit"))
 		return
 	}
 
+	backend, err := p.registry.GetBackend(getLoadTestType(r))
+	if err != nil {
+		render.Render(w, r, cHttp.ErrResponse(http.StatusBadRequest, err.Error()))
```
boekkooi-fresh, line 188: Can we add a logger call here to help debug?

pkg/proxy/proxy.go
```diff
@@ -181,12 +177,24 @@ func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	err = backend.TransformLoadTestSpec(&ltSpec)
+	if err != nil {
+		render.Render(w, r, cHttp.ErrResponse(http.StatusBadRequest, err.Error()))
```
boekkooi-fresh, line 194: Can we add a logger call here to help debug?

pkg/backends/options.go
```diff
@@ -1,19 +1,20 @@
...
 
 	"k8s.io/client-go/kubernetes"
 	coreListersV1 "k8s.io/client-go/listers/core/v1"
+
+	clientSetV "github.com/hellofresh/kangal/pkg/kubernetes/generated/clientset/versioned"
 )
 
 // Option allows to configure backends
-type Option func(*Registry)
+type Option func(*registry)
```
boekkooi-fresh, line 13: Why not use the interface?

pkg/proxy/api.go
```diff
@@ -68,9 +68,9 @@ func RunAPIServer(ctx context.Context, cfg Config, rr APIRunner) error {
 			return fmt.Errorf("could not create API TCP listener: %w", err)
 		}
 
-		rr.Logger.Info("Running gRPC server...", zap.String("addr", grpcAddress))
+		rr.Logger.Info("Running gRPC API server...", zap.String("addr", grpcAddress))
```
boekkooi-fresh, line 71: All changes in this file look like a merge conflict.

pkg/backends/options.go
diegomarangoni: all the options here are meant to work only with `registry` and any other implementation

pkg/backends/backend.go
```diff
@@ -2,75 +2,74 @@ package backends
...
-	CheckOrCreateResources(ctx context.Context) error
+// Backend is basic methods a backend must implement
+type Backend interface {
+	// Type must return backend unique type string, called by both Proxy and Controller
+	Type() loadTestV1.LoadTestType
 
-	// CheckOrUpdateLoadTestStatus check current LoadTest progress
-	CheckOrUpdateStatus(ctx context.Context) error
+	// TransformLoadTestSpec should validate and transform LoadTestSpec, called by Proxy
+	TransformLoadTestSpec(spec *loadTestV1.LoadTestSpec) error
```
aleksandrzhiliaev, line 20: What is the meaning of `TransformLoadTestSpec`? 
Does it mean that this method transforms spec to backend?

diegomarangoni: Means the backend has a chance to transform/validate the `LoadTestSpec` to ensure it matches his conditions

aleksandrzhiliaev:
+1

PR #77, opened 2020-11-30
EES-4643 Convert from Load Test to gRPC enum using map

vgarvardt:
Use map instead of switch to convert LT enums to gRPC enums. Maps to convert in other direction are initialised as well, but not used right now. Will be used in the upcoming PR with `Create` method implementation.


iammuho:
+1

PR #78, opened 2020-11-30
EES-4696 - Updating log levels for controller and proxy

s-radyuk:
This PR is dedicated to reducing the number of unnecessary logs:
- changes endpoint for health check from `/` to `/status`
- excludes log for requesting `/status` from info level
- excludes log for syncing load tests from info level


vgarvardt:
👍

PR #79, opened 2020-11-30
EES-4643 Implemented Create Load Test gRPC method

vgarvardt:
Probably the most complicated part because of file uploads that we use and that are not directly supported in gRPC REST gateway.

Had to copy-paste almost all the create logic from existing handler as the logic in most cases is sequence of:

```
result, err := doSomething(...)
if err != nil {
    renderer.Render(http.SomeErrorCode, "error message")
    return
}

### CODE REVIEW

File: pkg/proxy/service.go [hunk 1]
```diff
@@ -59,11 +80,165 @@ func (s *implLoadTestServiceServer) Get(ctx context.Context, in *grpcProxyV2.Get
 	}, nil
 }
 
+// Create creates new load test
+func (s *implLoadTestServiceServer) Create(ctx context.Context, in *grpcProxyV2.CreateRequest) (*grpcProxyV2.CreateResponse, error) {
```
diegomarangoni, line 84: Would be better to abstract all the logic inside this function and re-use on both v1/v2 endpoints?
Because now we have two completely different endpoints to maintain.

diegomarangoni: Same thing for the other endpoints

vgarvardt: I decsribed it in the PR description - it was my first intention, but the whole logic is just simple calls to something and then returning backend-agnostic result (different http errors) in case of the problem. Refactoring this may be separate task.

diegomarangoni:
+1

PR #85, opened 2020-12-02
EES-4643 Slightly simplified LT creation handler logic

vgarvardt:
Tiny code improvements:
- early return instead of `else` flow
- use type field from the spec that was already extracted from request instead of getting it from request again


diegomarangoni:
+1

PR #88, opened 2020-12-03
EES-4642 Moved gRPC tests from proxy to service

vgarvardt:
Tests for `Get` method were placed to proxy integration tests as it was much easier to kick off API. Now that we `Create` method implemented and covered with tests `Get` tests may join them.


s-radyuk:
+1

PR #89, opened 2020-12-04
EES-4642 Use typed phase and input validation in List request

vgarvardt:
To make input more strict and options reusable.


s-radyuk:
+1

PR #91, opened 2020-12-07
EES-4642 Default and max limit for LT lists

vgarvardt:
Do not allow unlimited lists by default.

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -34,16 +34,18 @@ var (
 // Proxy handler
 type Proxy struct {
 	maxLoadTestsRun int
+	maxListLimit    int64
 	registry        backends.Registry
 	kubeClient      *kube.Client
 }
 
 // NewProxy returns new Proxy handlers
-func NewProxy(maxLoadTestsRun int, registry backends.Registry, kubeClient *kube.Client) *Proxy {
+func NewProxy(maxLoadTestsRun int, registry backends.Registry, kubeClient *kube.Client, maxListLimit int64) *Proxy {
 	return &Proxy{
 		maxLoadTestsRun: maxLoadTestsRun,
 		registry:        registry,
 		kubeClient:      kubeClient,
+		maxListLimit:    maxListLimit,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


diegomarangoni:
Would be nice to have this specified on the documentation

pkg/proxy/proxy.go
```diff
@@ -34,16 +34,18 @@ var (
...
 type Proxy struct {
 	maxLoadTestsRun int
+	maxListLimit    int64
 	registry        backends.Registry
 	kubeClient      *kube.Client
 }
 
 // NewProxy returns new Proxy handlers
-func NewProxy(maxLoadTestsRun int, registry backends.Registry, kubeClient *kube.Client) *Proxy {
+func NewProxy(maxLoadTestsRun int, registry backends.Registry, kubeClient *kube.Client, maxListLimit int64) *Proxy {
```
diegomarangoni, line 43:
```suggestion
func NewProxy(maxLoadTestsRun int, maxListLimit int64, registry backends.Registry, kubeClient *kube.Client) *Proxy {
```
what do you think about grouping parameters with same name?

pkg/proxy/request.go
```diff
@@ -74,6 +76,10 @@ func fromHTTPRequestToListOptions(r *http.Request) (*kubernetes.ListOptions, err
 			return nil, err
 		}
 
+		if limit > maxListLimit {
+			return nil, fmt.Errorf("limit value is too big, max possible value is %d", maxListLimit)
```
s-radyuk, line 80: Let's add a test for this case to pkg/proxy/proxy_test.go

pkg/proxy/proxy.go
vgarvardt: I would group them by importance, tbh, but I did not want to change a lot - adding to the end is the simplest case.

pkg/proxy/request.go
vgarvardt: done

diegomarangoni:
Can you add this new env var to helm chart pls?

vgarvardt:
@diegomarangoni this change has nothing to do with helm chart. Currently we do not have docs with all the available env vars. Will create a task to create it.

diegomarangoni:
+1

PR #92, opened 2020-12-07
Wait for JMeter worker pods to be running before master creation

diegomarangoni:
EES-4737
Some load tests are randomly failing due the JMeter master pod being ready before workers pods.
This PR is intended to mitigate this by waiting each worker pod be on running state before proceeding.


s-radyuk:
+1

PR #93, opened 2020-12-08
EES-4642 Allow more than one test in gRPC integration tests

vgarvardt:
Increase of allowed tests will be used for `List` method tests as more than one test is required to cover pagination.


diegomarangoni:
+1

PR #95, opened 2020-12-10
Move helper package to backends

diegomarangoni:
EES-4269
Package `core/help` is too generic. This PR splits those packages into other more specific.


pkg/core/resources/resources.go
```diff
@@ -1,10 +1,6 @@
-package helper
+package resources
```
vgarvardt, line 1: As I see these `Resources` are used only in Backend, so does it make sense to leave it in the `core`? This thing has very specific purpose.

pkg/core/envs/envs.go
```diff
@@ -0,0 +1,33 @@
...
+	"strings"
+)
+
+var (
+	// ErrInvalidCSVFormat when the number of columns is different than two
+	ErrInvalidCSVFormat = errors.New("Invalid csv format, expecting: key, value")
+)
+
+// ReadEnvs reads data from csv file to save it as a map for creating a secret
+func ReadEnvs(envVars string) (map[string]string, error) {
```
vgarvardt, line 16: Same as with `Resources` - this is very specific piece of functionality used by only one piece of the app (backend), so IMO it should reside there. Core should be something general-purpose, reused in many parts of the app (like rqctx)

pkg/backends/jmeter/resources_test.go
```diff
@@ -3,11 +3,12 @@ package jmeter
...
-	loadtestv1 "github.com/hellofresh/kangal/pkg/kubernetes/apis/loadtest/v1"
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
+
+	"github.com/hellofresh/kangal/pkg/backends"
+	loadTestV1 "github.com/hellofresh/kangal/pkg/kubernetes/apis/loadtest/v1"
 )
 
 var logger = zap.NewNop()
```
vgarvardt, line 14: pls replace global logger instance with zaptest instances per test

vgarvardt:
👍

PR #97, opened 2020-12-17
EES-4392 Add created and finished loadtest metric

s4nji:
This PR:
- Adds two new metrics: `created_loadtests_count` and `finished_loadtests_count`
  - `created_loadtests_count` is incremented when a new loadtest is created
  - `finished_loadtests_count` is incremented when a loadtest phase is set to finished
- Adds error handling to `updateLoadTestStatus` in `syncHandler`
  - Errors here were returned, but ignored as it's only called with `defer` in `syncHandler`
- Renames `getLoadTestStatusFromJobs` to `determineLoadTestStatusFromJobs`

### DIFF
--- a/pkg/backends/fake/backend.go
+++ b/pkg/backends/fake/backend.go
@@ -115,7 +115,7 @@ func (b *Backend) SyncStatus(ctx context.Context, _ loadTestV1.LoadTest, loadTes
 
 	// Get Fake job in namespace and update the LoadTest status with
 	// the Job status
-	loadTestStatus.Phase = getLoadTestPhaseFromJob(job.Status)
+	loadTestStatus.Phase = determineLoadTestPhaseFromJob(job.Status)
 	loadTestStatus.JobStatus = job.Status
 
 	return nil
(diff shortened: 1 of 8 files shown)


pkg/backends/fake/backend.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/hellofresh/kangal/pkg/core/observability"
```
diegomarangoni, line 7: can you move this import to the section down below?

pkg/backends/jmeter/backend.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/hellofresh/kangal/pkg/core/observability"
```
diegomarangoni, line 9: same thing here

pkg/backends/jmeter/backend.go
```diff
@@ -214,6 +216,7 @@ func (b *Backend) Sync(ctx context.Context, loadTest loadTestV1.LoadTest, report
 			return err
 		}
 
+		stats.Record(ctx, observability.MCreatedLoadtestCountStat.M(1))
```
diegomarangoni, line 219: why not add this to `controller` instead of to every backend?

pkg/backends/locust/backend.go
```diff
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/hellofresh/kangal/pkg/core/observability"
```
diegomarangoni, line 8: import sort order again :)

diegomarangoni:
just realized we don't have an example grafana dashboard on the repo.
anyways, not sure if is related to this PR

s4nji:
> just realized we don't have an example grafana dashboard on the repo.
> anyways, not sure if is related to this PR

It will be done in a separate PR

pkg/backends/jmeter/backend.go
s4nji: Good catch, it's now incremented as new namespaces are created (f7c3d63)

diegomarangoni:
+1

PR #98, opened 2021-01-05
Add integration test for grpc-gateway list method with tags parameter

s4nji:
This PR adds integration test to ensure that `tags` query string parameter passed to `LIST` `/v2/load-test` endpoint in the grpc gateway is working.


vgarvardt:
👍

PR #99, opened 2021-01-13
Loadtest syncHandler: Stop processing and updating after loadtest deletion

s4nji:
**Current behavior:**  
After we delete a loadtest (due to staleness), we still attempt to update it, and then update its status (via deferred execution).  
This produces two error log lines as attempts to update the loadtest will fail (it has already been deleted).

**This PR:**  
Moves the check earlier, so we can stop further processing if a loadtest is to be deleted.

### DIFF
--- a/pkg/controller/loadtest.go
+++ b/pkg/controller/loadtest.go
@@ -323,12 +323,14 @@ func (c *Controller) syncHandler(key string) error {
 		// LoadTest has been deleted at this point, so we stop further processing.
 		return nil
 	}
-	
+
+	// get report url
 	var reportURL string
 	if c.cfg.KangalProxyURL != "" {
 		reportURL = fmt.Sprintf("%s/load-test/%s/report", c.cfg.KangalProxyURL, loadTest.GetName())
 	}
 
+	// get backend
 	backend, err := c.registry.GetBackend(loadTest.Spec.Type)
 	if err != nil {
 		return fmt.Errorf("failed to resolve backend: %w", err)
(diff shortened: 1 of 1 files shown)


pkg/controller/loadtest.go
```diff
@@ -303,18 +303,42 @@ func (c *Controller) syncHandler(key string) error {
 	}
 	// copy object before mutate it
 	loadTest := loadTestFromCache.DeepCopy()
-	defer c.updateLoadTestStatus(ctx, key, loadTest, loadTestFromCache)
 
+	// check and delete stale finished/errored loadtests
+	if checkLoadTestLifeTimeExceeded(loadTest, c.cfg.CleanUpThreshold) {
+		c.logger.Info("Deleting loadtest",
```
vgarvardt, line 309:
```suggestion
		c.logger.Info("Deleting loadtest due to exceeded lifetime",
```

pkg/controller/loadtest.go
```diff
@@ -303,18 +303,42 @@ func (c *Controller) syncHandler(key string) error {
...
+			zap.String("loadtest", loadTest.Name),
+			zap.String("phase", string(loadTest.Status.Phase)),
+		)
+		err = c.kangalClientSet.KangalV1().LoadTests().Delete(ctx, loadTest.Name, metaV1.DeleteOptions{})
+		if err != nil {
+			// The LoadTest resource may be conflicted, in which case we stop processing.
+			if errors.IsConflict(err) {
+				utilRuntime.HandleError(fmt.Errorf("there is a conflict with loadtest '%s' between datastore and cache. it might be because object has been removed or modified in the datastore", key))
```
vgarvardt, line 317:
```suggestion
				utilRuntime.HandleError(fmt.Errorf("there is a conflict with loadtest %q between datastore and cache. it might be because object has been removed or modified in the datastore", key))
```

vgarvardt:
👍

PR #100, opened 2021-01-13
Loadtest syncHandler: Don't update loadtest status with stale loadtest object

s4nji:
**Current behavior:**  
Kangal controller updates the loadtest resource, then updates its status again via a deferred call using an outdated loadtest object.  
This almost always produces an error since they're conflicted.

**This PR:**  
Updates deferred status update to use loadtest object with actually updated status.  
[This snippet](https://play.golang.org/p/lyhUp24cC54) showcases the difference between function-wrapped deferred calls.


s4nji:
Upon closer check this doesn't seem to remove reoccurring conflict log messages when updating status; it is still happening.  
Closing

PR #102, opened 2021-01-15
Implement gRPC Delete Method

s4nji:
This PR implements delete method in gRPC/REST proxy.

### DIFF
--- a/proto/grpc/proxy/v2/proxy.proto
+++ b/proto/grpc/proxy/v2/proxy.proto
@@ -2,7 +2,6 @@ syntax = "proto3";
 package grpc.proxy.v2;
 
 import "google/protobuf/duration.proto";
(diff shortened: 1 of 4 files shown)


proto/grpc/proxy/v2/proxy.proto
```diff
@@ -106,4 +111,10 @@ service LoadTestService {
             get: "/v2/load-test"
         };
     }
+
+    rpc Delete (DeleteRequest) returns (google.protobuf.Empty) {
```
vgarvardt, line 115: IMO it makes sense to use `DeleteResponse` type as result with empty fields list to make it extensible in case we'l need it.

s4nji: Done via 116dbce.

vgarvardt:
👍

PR #103, opened 2021-01-18
EES-4164 - moving csv check from controller to proxy

s-radyuk:
This PR adds a CSV file check to the proxy, which allows to validate and fail fast when the file is not a valid CSV.

1 - validating testData is a valid CSV file
2 - validating EnvVars is a valid CSV with `key,value` content
3 - loadTest CRD is updated to specify EnvVars as object (map[string]string), not a simple string.

Before both checks were performed on the controller side. Now moved to the proxy.


s4nji (approved):

diegomarangoni (approved):
nice improvement 👍🏻

diegomarangoni:
+1

PR #106, opened 2021-01-22
Apply timeout to kubeclient and remove KubeTimeout

s4nji:
This PR:
- Sets a default 5 seconds timeout to controller and proxy kubeClient 
  - Adds `KUBE_CLIENT_TIMEOUT` environment variable to configure this
- Sets a default 15 seconds timeout to kubeClients in integration tests
- Removes all uses of `KubeTimeout`
- Adds 60 seconds timeout to controller syncHandler
- Retains 30 seconds timeout (from `KubeTimeout`) when waiting for kube resources to appear

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -8,7 +8,6 @@ import (
 	"github.com/kelseyhightower/envconfig"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/hellofresh/kangal/pkg/core/observability"
 	kube "github.com/hellofresh/kangal/pkg/kubernetes"
@@ -48,7 +47,7 @@ func NewAPICmd(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not initialise Prometheus exporter: %w", err)
 			}
 
-			k8sConfig, err := clientcmd.BuildConfigFromFlags(opts.masterURL, opts.kubeConfig)
+			k8sConfig, err := buildKubeClientConfig(cfg.MasterURL, opts.kubeConfig, cfg.KubeClientTimeout)
 			if err != nil {
 				return fmt.Errorf("building config from flags: %w", err)
 			}
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/kelseyhightower/envconfig"
 	"github.com/spf13/cobra"
-	kubeClient "k8s.io/client-go/kubernetes"
+	kubernetesClient "k8s.io/client-go/kubernetes"
 
 	"github.com/hellofresh/kangal/pkg/core/observability"
 	"github.com/hellofresh/kangal/pkg/kubernetes"
@@ -57,7 +57,7 @@ func NewAPICmd(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("building kangal clientset: %w", err)
 			}
 
-			kubeClientSet, err := kubeClient.NewForConfig(k8sConfig)
+			kubeClientSet, err := kubernetesClient.NewForConfig(k8sConfig)
 			if err != nil {
 				return fmt.Errorf("building kubernetes clientset: %w", err)
 			}
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -47,7 +47,7 @@ func NewAPICmd(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not initialise Prometheus exporter: %w", err)
 			}
 
-			k8sConfig, err := kubernetes.BuildClientConfig(cfg.MasterURL, opts.kubeConfig, cfg.KubeClientTimeout)
+			k8sConfig, err := kubernetes.BuildClientConfig(opts.masterURL, opts.kubeConfig, cfg.KubeClientTimeout)
 			if err != nil {
 				return fmt.Errorf("building config from flags: %w", err)
 			}
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -7,10 +7,10 @@ import (
 
 	"github.com/kelseyhightower/envconfig"
 	"github.com/spf13/cobra"
-	"k8s.io/client-go/kubernetes"
+	kubeClient "k8s.io/client-go/kubernetes"
 
 	"github.com/hellofresh/kangal/pkg/core/observability"
-	kube "github.com/hellofresh/kangal/pkg/kubernetes"
+	"github.com/hellofresh/kangal/pkg/kubernetes"
 	loadTestV1 "github.com/hellofresh/kangal/pkg/kubernetes/generated/clientset/versioned/typed/loadtest/v1"
 	"github.com/hellofresh/kangal/pkg/proxy"
 	"github.com/hellofresh/kangal/pkg/report"
@@ -47,7 +47,7 @@ func NewAPICmd(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not initialise Prometheus exporter: %w", err)
 			}
 
-			k8sConfig, err := buildKubeClientConfig(cfg.MasterURL, opts.kubeConfig, cfg.KubeClientTimeout)
+			k8sConfig, err := kubernetes.BuildKubeClientConfig(cfg.MasterURL, opts.kubeConfig, cfg.KubeClientTimeout)
 			if err != nil {
 				return fmt.Errorf("building config from flags: %w", err)
 			}
@@ -57,13 +57,13 @@ func NewAPICmd(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("building kangal clientset: %w", err)
 			}
 
-			kubeClientSet, err := kubernetes.NewForConfig(k8sConfig)
+			kubeClientSet, err := kubeClient.NewForConfig(k8sConfig)
 			if err != nil {
 				return fmt.Errorf("building kubernetes clientset: %w", err)
 			}
 
 			loadTestClient := kangalClientSet.LoadTests()
-			kubeClient := kube.NewClient(loadTestClient, kubeClientSet, logger)
+			kubeClient := kubernetes.NewClient(loadTestClient, kubeClientSet, logger)
 
 			err = report.InitObjectStorageClient(cfg.Report)
 			if err != nil {
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -47,7 +47,7 @@ func NewAPICmd(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not initialise Prometheus exporter: %w", err)
 			}
 
-			k8sConfig, err := kubernetes.BuildKubeClientConfig(cfg.MasterURL, opts.kubeConfig, cfg.KubeClientTimeout)
+			k8sConfig, err := kubernetes.BuildClientConfig(cfg.MasterURL, opts.kubeConfig, cfg.KubeClientTimeout)
 			if err != nil {
 				return fmt.Errorf("building config from flags: %w", err)
 			}
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -54,7 +54,7 @@ func NewControllerCmd() *cobra.Command {
 				return err
 			}
 
-			kubeCfg, err := kubernetes.BuildKubeClientConfig(cfg.MasterURL, cfg.KubeConfig, cfg.KubeClientTimeout)
+			kubeCfg, err := kubernetes.BuildClientConfig(cfg.MasterURL, cfg.KubeConfig, cfg.KubeClientTimeout)
 			if err != nil {
 				return fmt.Errorf("error building kubeConfig: %w", err)
 			}
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -58,7 +58,7 @@ func NewControllerCmd() *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("error building kubeConfig: %w", err)
 			}
-			kubeCfg.Timeout = time.Duration(cfg.KubeClientTimeoutSeconds) * time.Second
+			kubeCfg.Timeout = cfg.KubeClientTimeout
 
 			kubeClient, err := kubernetes.NewForConfig(kubeCfg)
 			if err != nil {
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/client-go/rest"
+
 	"github.com/kelseyhightower/envconfig"
 	"github.com/spf13/cobra"
 	kubeInformers "k8s.io/client-go/informers"
@@ -54,11 +56,10 @@ func NewControllerCmd() *cobra.Command {
 				return err
 			}
 
-			kubeCfg, err := clientcmd.BuildConfigFromFlags(cfg.MasterURL, cfg.KubeConfig)
+			kubeCfg, err := buildKubeClientConfig(cfg.MasterURL, cfg.KubeConfig, cfg.KubeClientTimeout)
 			if err != nil {
 				return fmt.Errorf("error building kubeConfig: %w", err)
 			}
-			kubeCfg.Timeout = cfg.KubeClientTimeout
 
 			kubeClient, err := kubernetes.NewForConfig(kubeCfg)
 			if err != nil {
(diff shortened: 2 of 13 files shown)


pkg/controller/config.go
```diff
@@ -11,12 +11,17 @@ type Config struct {
...
+
 	// CleanUpThresholdEnvVar is used if we want to increase the amount of time a
 	// load test lives for, the default is 1 hour. (ex. 5h)
 	CleanUpThreshold time.Duration `envconfig:"CLEANUP_THRESHOLD" default:"1h"`
+
 	// S3 compatible configuration access keys and endpoints needed to store load test reports
 	KangalProxyURL string `envconfig:"KANGAL_PROXY_URL" default:""`
 
+	// KubeClientTimeoutSeconds specifies timeout for each operation done by kube client
+	KubeClientTimeoutSeconds uint `envconfig:"KUBE_CLIENT_TIMEOUT_SECONDS" default:"5"`
```
vgarvardt, line 23: is there a raeson not to use `time.Duration` instead of seconds? `envconfig` supports this type out of the box

pkg/controller/loadtest.go
```diff
@@ -275,7 +274,7 @@ func (c *Controller) processNextWorkItem() bool {
 // converge the two. It then updates the Status block of the LoadTest resource
 // with the current status of the resource.
 func (c *Controller) syncHandler(key string) error {
-	ctx, cancel := context.WithTimeout(context.Background(), kubekangal.KubeTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
vgarvardt, line 277: Why do we have fixed time here now? Seems like a magic const now.

pkg/controller/config.go
s4nji: Oh! I didn't know about that, will use `time.Duration`

pkg/controller/loadtest.go
s4nji: It should never have used "kube timeout" in the first place, so we can either remove it or set a new duration for it.  
It can be labelled as `syncTimeout` constant if that helps.

pkg/controller/loadtest.go
```diff
@@ -36,6 +35,7 @@ const (
 	controllerAgentName = "kangal"
 	falseString         = "false"
 	trueString          = "true"
+	syncHandlerTimeout  = time.Minute
```
aleksandrzhiliaev, line 38: Not needed to make this configurable?

cmd/controller.go
```diff
@@ -58,6 +58,7 @@ func NewControllerCmd() *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("error building kubeConfig: %w", err)
 			}
+			kubeCfg.Timeout = cfg.KubeClientTimeout
```
boekkooi-fresh, line 61: Why is `cfg.KubeClientTimeout` not passed into `clientcmd.BuildConfigFromFlags`? That way we avoid people to forget assigning `kubeCfg.Timeout` (same question for [cmd/proxy.go line 55](https://github.com/hellofresh/kangal/pull/106/files#diff-c255784f7c4412e83a26aa04ec6156922b0a0bec2e2ff55e10f8981346da65a5R55))

pkg/core/waitfor/waitfor.go
```diff
@@ -53,7 +53,7 @@ func (Condition) LoadTestFinished(event watch.Event) (bool, error) {
 
 // Resource waits until a kubernetes resources to match a condition
 func Resource(obj watch.Interface, condFunc watchtools.ConditionFunc) (*watch.Event, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), kubernetes.KubeTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
aleksandrzhiliaev, line 56: Same for `30` - is it fine to have magic const here or maybe move it to configuration?

pkg/controller/jmeter_integration_test.go
```diff
@@ -162,6 +162,7 @@ func kubeTestClient() clientSetV.Clientset {
 	}
 
 	config, err := BuildConfig()
+	config.Timeout = 60 * time.Second
```
boekkooi-fresh, line 165: Why is this not set within `BuildConfig`? Seems like this could fix some issues with people forgetting to set the timeout.

pkg/controller/loadtest.go
```diff
@@ -275,7 +275,7 @@ func (c *Controller) processNextWorkItem() bool {
 // converge the two. It then updates the Status block of the LoadTest resource
 // with the current status of the resource.
 func (c *Controller) syncHandler(key string) error {
-	ctx, cancel := context.WithTimeout(context.Background(), kubekangal.KubeTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), syncHandlerTimeout)
```
boekkooi-fresh, line 278: Hard coding this could cause a similar issue that you are trying to fix with `KUBE_CLIENT_TIMEOUT` maybe add a to do to introduce a `KUBE_WAIT_FOR_TIMEOUT` or something?

Also maybe the split of `kubekangal.KubeTimeout` should be split into 2 PR's?

pkg/controller/test_helper.go
```diff
@@ -215,5 +205,6 @@ func BuildConfig() (*rest.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+	config.Timeout = 30 * time.Second
```
boekkooi-fresh, line 208: Possibly move this to a `const` so someone knows what this timeout is configuring.

pkg/controller/loadtest.go
s4nji: No; if sync takes more than one minute then something is very wrong, and increasing it would not be a good fix.

pkg/core/waitfor/waitfor.go
boekkooi-fresh, line 56: Hard coding this could cause a similar issue that you are trying to fix with `KUBE_CLIENT_TIMEOUT` maybe add a to do to introduce a `KUBE_WAIT_FOR_TIMEOUT` or something?

pkg/controller/jmeter_integration_test.go
s4nji: Apparently this is already set in `BuildConfig`, so this can be removed

cmd/controller.go
s4nji: I don't understand, could you explain more? 🤔 

From what I understand, [`clientcmd.BuildConfigFromFlags` only takes path to `kubeconfig` file](https://github.com/kubernetes/kubernetes/blob/3433f099fb0b5c51d3b549372364f818242f7312/staging/src/k8s.io/client-go/tools/clientcmd/client_config.go#L613-L625), in which no timeout can be set to be used as the rest timeout config.

pkg/controller/loadtest.go
s4nji: Partially explained in https://github.com/hellofresh/kangal/pull/106#discussion_r562691500; the entire `syncHandler` should not be wrapped with a 30 seconds "kube timeout". It was a somewhat lazy way to prevent kube timeout, which worked until the entire sync process took longer than 30 seconds.  
This is now done by adding 5 seconds timeout to each individual operation done by `kubeClient`.

I previously considered removing `syncHandlerTimeout` since the individual timeouts should be sufficient, but I added a 60 seconds timeout in case it does get too long. Now I think the better approach is perhaps to add a metric to track the time taken for the sync process (and have alerts if it's taking too long).

cmd/controller.go
boekkooi-fresh: A I didn't realize that `BuildConfigFromFlags` is part of k8s.
In that case why don't we create a function `BuildConfigFromFlags` within k8s where we require the `Timeout` as an argument? This would avoid all these `kubeCfg.Timeout =` all over the place :smile:

pkg/core/waitfor/waitfor.go
s4nji: Added in ae26927

pkg/controller/jmeter_integration_test.go
s4nji: Done in 6b6ec81

pkg/core/waitfor/waitfor.go
s4nji: Done in ae26927

pkg/controller/test_helper.go
s4nji: Done in 6b6ec81

cmd/controller.go
s4nji: Done in 68ee8a2

cmd/controller.go
```diff
@@ -131,3 +133,13 @@ func convertAnnotationToMap(s []string) (map[string]string, error) {
 	}
 	return m, nil
 }
+
+func buildKubeClientConfig(masterURL string, kubeConfigPath string, timeout time.Duration) (*rest.Config, error) {
```
boekkooi-fresh, line 137: I'm not sure but would placing this inside the [`kubernetes`](https://github.com/hellofresh/kangal/tree/master/pkg/kubernetes) package not be cleaner?

cmd/api.go
```diff
@@ -48,7 +47,7 @@ func NewAPICmd(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not initialise Prometheus exporter: %w", err)
 			}
 
-			k8sConfig, err := clientcmd.BuildConfigFromFlags(opts.masterURL, opts.kubeConfig)
+			k8sConfig, err := buildKubeClientConfig(cfg.MasterURL, opts.kubeConfig, cfg.KubeClientTimeout)
```
boekkooi-fresh, line 50: More for consistency but why is `kubeConfig` private and are `MasterURL` and `KubeClientTimeout` not?

pkg/controller/loadtest.go
boekkooi-fresh: Looking at this function and the amount of calls it is doing to `cache`, `kubernetes`, `backend` I think making this configurable or at least based on something configurable maybe best.
For example let's say you set kubeClientTimeout to 2 minutes then this sync will now always start failing assuming k8s responds after 60s. 

It is probably best to allow a user to specify this timeout or to calculate it based on timeouts set for k8s, backend, cache etc.

cmd/api.go
s4nji: The `opts` are not used anywhere else outside the `cmd` package.
We can move it to a separate package to make them all public and consistent but that seems like more trouble than it's worth.

pkg/controller/loadtest.go
s4nji: It is now an environment variable (fd31a83)

cmd/controller.go
s4nji: Done via 8aad2bc

cmd/api.go
boekkooi-fresh: I think we have a misunderstanding here, let my try a different angle.
Could you explain why you made `MasterURL` public? Can you also explain why `KubeClientTimeout` was created as public.

cmd/controller.go
```diff
@@ -54,12 +54,12 @@ func NewControllerCmd() *cobra.Command {
...
 			}
 
-			kubeCfg, err := clientcmd.BuildConfigFromFlags(cfg.MasterURL, cfg.KubeConfig)
+			kubeCfg, err := kubernetes.BuildKubeClientConfig(cfg.MasterURL, cfg.KubeConfig, cfg.KubeClientTimeout)
 			if err != nil {
 				return fmt.Errorf("error building kubeConfig: %w", err)
 			}
 
-			kubeClient, err := kubernetes.NewForConfig(kubeCfg)
+			kubeClient, err := kubeClient.NewForConfig(kubeCfg)
```
boekkooi-fresh, line 62: Maybe it's better to rename the alias `kubeClient` to `kubernetesClient`? This way you don't have a variable and import named `kubeClient`

cmd/proxy.go
```diff
@@ -58,13 +57,13 @@ func NewProxyCmd() *cobra.Command {
...
 
-			kubeClientSet, err := kubernetes.NewForConfig(k8sConfig)
+			kubeClientSet, err := kubeClient.NewForConfig(k8sConfig)
 			if err != nil {
 				return fmt.Errorf("building kubernetes clientset: %w", err)
 			}
 
 			loadTestClient := kangalClientSet.LoadTests()
-			kubeClient := kube.NewClient(loadTestClient, kubeClientSet, logger)
+			kubeClient := kubernetes.NewClient(loadTestClient, kubeClientSet, logger)
```
boekkooi-fresh, line 66: Maybe it's better to rename the alias `kubeClient` to `kubernetesClient`? This way you don't have a variable and import named `kubeClient`

pkg/controller/config.go
```diff
@@ -11,12 +11,20 @@ type Config struct {
...
 	CleanUpThreshold time.Duration `envconfig:"CLEANUP_THRESHOLD" default:"1h"`
+
 	// S3 compatible configuration access keys and endpoints needed to store load test reports
 	KangalProxyURL string `envconfig:"KANGAL_PROXY_URL" default:""`
 
+	// KubeClientTimeout specifies timeout for each operation done by kube client
+	KubeClientTimeout time.Duration `envconfig:"KUBE_CLIENT_TIMEOUT" default:"5s"`
+
+	// SyncHandlerTimeout specifies the time limit for each sync operation
+	SyncHandlerTimeout time.Duration `envconfig:"SYNC_HANDLER_TIMEOUT" default:"60s"`
```
boekkooi-fresh, line 26: What do you thing about logging a warning/error when `SyncHandlerTimeout` is less then `KubeClientTimeout`?

pkg/kubernetes/client.go
```diff
@@ -240,3 +236,14 @@ func sortWorkerPods(pods *coreV1.PodList) {
 		return strings.Compare(pods.Items[i].ObjectMeta.Name, pods.Items[j].ObjectMeta.Name) < 0
 	})
 }
+
+// BuildKubeClientConfig is used in cmd package
+func BuildKubeClientConfig(masterURL string, kubeConfigPath string, timeout time.Duration) (*restClient.Config, error) {
```
boekkooi-fresh, line 241: Maybe rename this to `BuildClientConfig` since `Kube` is already part of the package this func exists in.

cmd/api.go
s4nji: > Could you explain why you made `MasterURL` public?

Seems like I got confused here (and changed `opts.masterURL` to `cfg.MasterURL` since they were already inconsistent([api], [controller], and [proxy]) in the first place. I reverted it to use `opts.masterURL` again in 5e7ef06.  
This and several other items from config can be moved to `opts`, but I think it should be done in another pull-request.

> Can you also explain why `KubeClientTimeout` was created as public.

For consistency; both `controller.Config` and `proxy.Config` are where envconfigs are located. I suppose we can add/create `cmd.Config` and move it there, but it feels too much for one envconfig.

[api]: https://github.com/hellofresh/kangal/blob/ee235493c32b20adeded3cf69d678e9525903d70/cmd/api.go#L51
[controller]: https://github.com/hellofresh/kangal/blob/ee235493c32b20adeded3cf69d678e9525903d70/cmd/controller.go#L57
[proxy]: https://github.com/hellofresh/kangal/blob/ee235493c32b20adeded3cf69d678e9525903d70/cmd/proxy.go#L51

cmd/controller.go
s4nji: Done via 97b1771

cmd/proxy.go
s4nji: Done via 97b1771

pkg/kubernetes/client.go
s4nji: Done via 97b1771

pkg/controller/config.go
s4nji: I think it would be helpful, but should be done in another pull-request (perhaps to also include checks/warnings for other configs) so it can be planned/reviewed better.

boekkooi-fresh: :+1: Can you please create a ticket for this

pkg/controller/loadtest.go
s4nji: This topic is continued in https://github.com/hellofresh/kangal/pull/106#discussion_r563588998

boekkooi-fresh (approved):

boekkooi-fresh:
:+1:

PR #107, opened 2021-02-01
Remove loadtest update

s4nji:
Loadtest object needed to updated because loadtest spec was being updated, but no longer being done as of #48.

This PR removes the unnecessary update.


diegomarangoni (approved):

diegomarangoni:
+1

PR #109, opened 2021-02-03
Improve Logging

s4nji:
This PR:
- Updates logs to include loadtest name field consistently
- Adds logging when status is updated
- Renames log fields to be more consistent (mainly `loadtest` casing)


diegomarangoni (approved):
nice improvement, let's merge!

PR #110, opened 2021-02-03
Implement gRPC GetLogs method

s4nji:
This PR (re)implements GetLogs endpoint in the proxy in grpc.


diegomarangoni:
should we use streams instead?

s4nji:
> should we use streams instead?

It's currently not possible to do streaming response with `grpc-gateway` (see `https://github.com/grpc-ecosystem/grpc-gateway/issues/958`)

vgarvardt:
I think we can close it and gradually remove this experimental API - it gave a lot of experience and knowledge, but, unfortunately, it does not fit the project model. We should get back to generating OpenAPI from code approach IMO. This time with the other tools, probably start with https://github.com/swaggest/swgen

PR #118, opened 2021-04-15
Fix non tar static report serving

### LINKED ISSUE #117: Locust Report - HTTP 404
I was able to run the locust test successfully using Kangal. But when I query the report API, I am getting `404 page not found`. 

```
curl http://127.0.0.1:8081/load-test/loadtest-terrific-chinchilla/report/
```

Error:
```
404 page not found
```

diegomarangoni:
closes #117

Non-tar static files were unable to be served because they didn't have a dedicated condition for those kinds of files.
This PR adds a condition to handle non-tar files.

### DIFF
--- a/pkg/report/handler.go
+++ b/pkg/report/handler.go
@@ -58,8 +58,21 @@ func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+
 		objStat, err := obj.Stat()
-		if nil == err && "application/x-tar" == objStat.ContentType {
+		// not found error can be a directory
+		if objErr, _ := err.(minio.ErrorResponse); http.StatusNotFound == objErr.StatusCode {
+			http.FileServer(fs).ServeHTTP(w, r)
+			return
+		}
+		// unknown error
+		if nil != err {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


pkg/report/handler.go
```diff
@@ -58,8 +58,21 @@ func ShowHandler() func(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+
 		objStat, err := obj.Stat()
-		if nil == err && "application/x-tar" == objStat.ContentType {
+		// not found error can be a directory
+		if objErr, _ := err.(minio.ErrorResponse); http.StatusNotFound == objErr.StatusCode {
```
iammuho, line 64: Why we are not returning the error here although it couldn't be found but still serving the file? Serving won't break because the file doesn't exist?

diegomarangoni: >Why we are not returning the error here although it couldn't be found but still serving the file? Serving won't break because the file doesn't exist?

Not really, `http.FileServer(fs)` will look for `index.html` inside the directory (if is actually a directory) and proper handler in case not find anything

>Regarding to your comment, shouldn't we check if request is a directory ?

There is no way to check if is a directory or not, actually, S3 does not have directory concept at all.

iammuho (approved):

vgarvardt (approved):

PR #123, opened 2021-04-21
Update Chart installation proccess

diegomarangoni:
The intention of this PR is to make chart installation a little easier and clearer, so it contains some documentation stuff along with default values for parameters and other stuff.

This PR introduces the following changes to the Helm chart:

- Update chart to helm 3
- Add automatic CRD install
- Remove Service `-k8s` suffix
- Remove ServiceMonitor `-k8s` suffix
- Update k8s Ingress resource
- Allows to helm install without set arguments
- Make docs kangal proxy url consistent
- Make example URLs more consistent
- Unquote yaml strings fields on values file


charts/kangal/Chart.yaml
```diff
@@ -1,12 +1,12 @@
...
+apiVersion: v2
 description: Run JMeter performance tests in Kubernetes cluster with Kangal
 keywords:
   - kubernetes
   - jmeter
   - performance tests
   - tests runner
 name: kangal
-version: 1.4.10
+version: 1.5.0
```
vgarvardt, line 9: I would say it should be major version bump, because of the `apiVersion: v2` and some breaking changes in templates

README.md
```diff
@@ -89,13 +89,7 @@ The component is responsible for managing all the aspects of the performance tes
...
 
 ### Installing using helm
-First, let's create the Custom Resource Definition by running:
-
-```shell
-$ kubectl apply -f https://raw.githubusercontent.com/hellofresh/kangal/master/charts/kangal/crd.yaml
-```
-
-Add the repository to Helm:
+First, add the repository to Helm:
```
ElenaForester, line 92: does helm install CRDs now? IIRC there were some difficulties with it

diegomarangoni: yes, just leave the CRD inside a folder called `crds` and helm will install if not exists

the difficulties are that helm does not support deletion or update

vgarvardt (approved):

PR #125, opened 2021-04-23
Make install text a bit clearer

diegomarangoni:
Improves the install docs text and make it clear, also adds a link to `values.yaml`.


diegomarangoni:
Using yaml `&version` thing is not working with helm for some reason... closing it

PR #127, opened 2021-05-03
Add AWS secure env

### LINKED ISSUE #126: Allow secure access to S3
Firstly, thank you for this project!

The MinIO client, configured for report uploading, is initialised with `secure` set to `false` (in the [`NewWithCredentials`](https://pkg.go.dev/gopkg.in/minio/minio-go.v6#NewWithCredentials) method):

https://github.com/hellofresh/kangal/blob/37a89a3569a310c2a501478fd8b0c055ad436293/pkg/report/report.go#L59

Being able to enable encryption in transit would be a good option to have from a [security perspective](https://docs.aws.amazon.com/AmazonS3/latest/userguide/security-best-practices.html#transit).

Perhaps this could be implemented in a backwards compatible manner, so the default is `secure` set to `false` but this could be set to `true` using configuration?

diegomarangoni:
fixes #126

This PR adds a new ENV called `AWS_USE_HTTPS` that allows controlling the use of HTTPS.

### DIFF
--- a/pkg/report/config.go
+++ b/pkg/report/config.go
@@ -8,6 +8,6 @@ type Config struct {
 	AWSRegion           string `envconfig:"AWS_DEFAULT_REGION" default:""`
 	AWSEndpointURL      string `envconfig:"AWS_ENDPOINT_URL" default:""`
 	AWSBucketName       string `envconfig:"AWS_BUCKET_NAME" default:""`
-	AWSUseHTTPS         string `envconfig:"AWS_USE_HTTPS" default:""`
+	AWSUseHTTPS         bool   `envconfig:"AWS_USE_HTTPS" default:"false"`
 	AWSPresignedExpires string `envconfig:"AWS_PRESIGNED_EXPIRES" default:""`
 }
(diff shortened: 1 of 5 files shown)


pkg/report/config.go
```diff
@@ -8,5 +8,6 @@ type Config struct {
 	AWSRegion           string `envconfig:"AWS_DEFAULT_REGION" default:""`
 	AWSEndpointURL      string `envconfig:"AWS_ENDPOINT_URL" default:""`
 	AWSBucketName       string `envconfig:"AWS_BUCKET_NAME" default:""`
+	AWSUseHTTPS         string `envconfig:"AWS_USE_HTTPS" default:""`
```
vgarvardt, line 11:
```suggestion
	AWSUseHTTPS         bool `envconfig:"AWS_USE_HTTPS" default:"false"`
```

pkg/report/report.go
```diff
@@ -55,8 +56,16 @@ func InitObjectStorageClient(cfg Config) error {
 	}
 	creds := credentials.NewChainCredentials(awsCredProviders)
 
+	secure := false
+	if cfg.AWSUseHTTPS != "" {
+		secure, err = strconv.ParseBool(cfg.AWSUseHTTPS)
```
vgarvardt, line 61: `envconfig` does this for you

aleksandrzhiliaev (approved):

PR #134, opened 2021-05-19
Fixing the process of loadtests deletion 

s-radyuk:
This PR: 
1. Updating controller integration test with a check if loadtest was deleted after finished. Setting `CLEANUP_THRESHOLD=10s` and `SYNC_HANDLER_TIMEOUT=10s` for test env to run Sync every 10s and delete the loadtest quickly.
2. Refactor controller integration test since all the checks depends on each other and it doesn't make sense to run it in subtest.
2. We need to run `checkLoadTestLifeTimeExceeded()` after `backend.SyncStatus()`. `SyncStatus` updates the loadtest status from jobs statuses. LoadTestStatus.JobStatus is empty until we run SyncStatus and can't be compared with `CleanUpThreshold` time.

### DIFF
--- a/pkg/controller/controller_integration_test.go
+++ b/pkg/controller/controller_integration_test.go
@@ -17,9 +17,10 @@ import (
 )
 
 func TestIntegrationKangalController(t *testing.T) {
-	// This integration test cover main idea and logic about Kangal controller
-	// First of all it creates new LoadTest resource, then it expects that Kangal Controller created resources
-	// and changed LoadTest status to "finished".
+	// This integration test covers main workflow of Kangal controller.
+	// First of all test_helper creates a new LoadTest object.
+	// After this Kangal controller creates all associated resources
+	// and updates LoadTest statuses from "created" to "finished".
 	if testing.Short() {
 		t.Skip("skipping test in short mode")
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/controller_integration_test.go
```diff
@@ -41,62 +43,60 @@ func TestIntegrationKangalController(t *testing.T) {
...
 	})
+	watchEvent, err = waitfor.Resource(watchObj, (waitfor.Condition{}).LoadTestFinished)
+	require.NoError(t, err)
+	loadtest = watchEvent.Object.(*loadTestV1.LoadTest)
+	require.Equal(t, loadTestV1.LoadTestFinished, loadtest.Status.Phase)
+
+	// Checking finished loadtest is deleted
+	// SyncHandler runs every 10s for integration test.
+	// We expect SyncHandler to delete finished loadtest after 10s but wait 30s to be sure
+	time.Sleep(30 * time.Second)
```
vgarvardt, line 97: Just to slightly speed up the test - probably it makes sense to run this check in the loop 3 times with 10s interval, or 6 times with 5s and fail only when after several times the test still persists

pkg/controller/controller_integration_test.go
```diff
@@ -41,62 +43,66 @@ func TestIntegrationKangalController(t *testing.T) {
...
+	for i := 0; i < 6; i++ {
+		time.Sleep(5 * time.Second)
+		lt, _ := clientSet.KangalV1().LoadTests().Get(context.Background(), expectedLoadtestName, metaV1.GetOptions{})
+		// assert that the returned object is empty which means lt "loadtest-fake-integration" was deleted
+		if lt.Name == "" && lt.Namespace == "" {
+			deleted = true
+			break
+		}
+	}
+	assert.True(t, deleted)
```
vgarvardt, line 107:
```suggestion
	assert.True(t, deleted, "Looks like test was not deleted during expected time frame")
```

pkg/controller/loadtest.go
```diff
@@ -523,3 +510,17 @@ func checkLoadTestLifeTimeExceeded(loadTest *loadTestV1.LoadTest, deleteThreshol
...
+	err := c.kangalClientSet.KangalV1().LoadTests().Delete(ctx, loadTest.Name, metaV1.DeleteOptions{})
+	if err != nil {
+		// The LoadTest resource may be conflicted, in which case we stop processing.
+		if errors.IsConflict(err) {
+			c.logger.Error("There is a conflict while deleting the loadtest", zap.Error(err))
+			utilRuntime.HandleError(fmt.Errorf("there is a conflict with loadtest %q between datastore and cache. it might be because object has been removed or modified in the datastore", key))
+			return
+		}
+		c.logger.Error("Failed to delete loadtest:", zap.Error(err))
+		return
```
lucasmdrs, line 524: There's no need for this return statement but everything else looks fine :+1:

pkg/controller/loadtest.go
```diff
@@ -523,3 +510,17 @@ func checkLoadTestLifeTimeExceeded(loadTest *loadTestV1.LoadTest, deleteThreshol
 
 	return false
 }
+
+func (c *Controller) deleteLoadTest(ctx context.Context, key string, loadTest *loadTestV1.LoadTest) {
+	err := c.kangalClientSet.KangalV1().LoadTests().Delete(ctx, loadTest.Name, metaV1.DeleteOptions{})
+	if err != nil {
```
vgarvardt, line 516: reducing nesting for the main logic may slightly improve readability, so this condition may be changed to

```go
if err == nil {
  return
}

// here comes the main logic on the top level

File: pkg/controller/loadtest.go [hunk 5]
```diff
@@ -523,3 +510,17 @@ func checkLoadTestLifeTimeExceeded(loadTest *loadTestV1.LoadTest, deleteThreshol
...
+		// The LoadTest resource may be conflicted, in which case we stop processing.
+		if errors.IsConflict(err) {
+			c.logger.Error("There is a conflict while deleting the loadtest", zap.Error(err))
+			utilRuntime.HandleError(fmt.Errorf("there is a conflict with loadtest %q between datastore and cache. it might be because object has been removed or modified in the datastore", key))
+			return
+		}
+		c.logger.Error("Failed to delete loadtest:", zap.Error(err))
```
vgarvardt, line 523:
```suggestion
		c.logger.Error("Failed to delete loadtest", zap.Error(err))
```

pkg/controller/loadtest.go
```diff
@@ -523,3 +510,17 @@ func checkLoadTestLifeTimeExceeded(loadTest *loadTestV1.LoadTest, deleteThreshol
...
 	return false
 }
+
+func (c *Controller) deleteLoadTest(ctx context.Context, key string, loadTest *loadTestV1.LoadTest) {
+	err := c.kangalClientSet.KangalV1().LoadTests().Delete(ctx, loadTest.Name, metaV1.DeleteOptions{})
+	if err != nil {
+		// The LoadTest resource may be conflicted, in which case we stop processing.
+		if errors.IsConflict(err) {
+			c.logger.Error("There is a conflict while deleting the loadtest", zap.Error(err))
+			utilRuntime.HandleError(fmt.Errorf("there is a conflict with loadtest %q between datastore and cache. it might be because object has been removed or modified in the datastore", key))
```
vgarvardt, line 520:
```suggestion
			utilRuntime.HandleError(fmt.Errorf("there is a conflict with loadtest %q between datastore and cache. It might be because object has been removed or modified in the datastore", key))
```

pkg/controller/loadtest.go
vgarvardt: now it is early return in case of no error

vgarvardt (approved):

lucasmdrs (approved):

PR #137, opened 2021-05-28
Feature/ees 5842 remove jmeter properties

s-radyuk:
This PR:
- removes the creation of jmeter.properties configmap, since it's now included in kangal-jmeter image [in this PR](https://github.com/hellofresh/kangal-jmeter/pull/17)
- updates logs level and the message in case of empty loadtest.spec.MasterConfig and loadtest.spec.WorkerConfig
- refactor SetDefaults() to set Image and Tag defaults separately


pkg/backends/jmeter/resources.go
```diff
@@ -180,7 +145,7 @@ func (b *Backend) NewPod(loadTest loadTestV1.LoadTest, i int, configMap *coreV1.
 	imageRef := fmt.Sprintf("%s:%s", loadTest.Spec.WorkerConfig.Image, loadTest.Spec.WorkerConfig.Tag)
 	if "" == loadTest.Spec.WorkerConfig.Image || "" == loadTest.Spec.WorkerConfig.Tag {
 		imageRef = fmt.Sprintf("%s:%s", b.workerConfig.Image, b.workerConfig.Tag)
-		logger.Warn("Loadtest.Spec.WorkerConfig is empty; using default worker image", zap.String("imageRef", imageRef))
+		logger.Debug("Loadtest.Spec.WorkerConfig is empty; using worker image from config", zap.String("imageRef", imageRef))
```
vgarvardt, line 148: maybe info would be better here and below?

diegomarangoni (approved):

PR #138, opened 2021-05-31
fixing the logic of Locust loadtest delition and adding more tests

s-radyuk:
This PR:
- updates the Locust backend to improve the logic of updating loadtest state 
- adds more unit tests for Locust backend

### DIFF
--- a/pkg/backends/locust/backend.go
+++ b/pkg/backends/locust/backend.go
@@ -206,8 +206,7 @@ func (b *Backend) SyncStatus(ctx context.Context, loadTest loadTestV1.LoadTest, 
 		loadTestStatus.Phase = loadTestV1.LoadTestCreating
 	}
 
-	if loadTestStatus.Phase == loadTestV1.LoadTestErrored ||
-		loadTestStatus.Phase == loadTestV1.LoadTestFinished {
+	if loadTestStatus.Phase == loadTestV1.LoadTestErrored {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/backends/locust/backend.go
```diff
@@ -206,8 +206,7 @@ func (b *Backend) SyncStatus(ctx context.Context, loadTest loadTestV1.LoadTest,
 		loadTestStatus.Phase = loadTestV1.LoadTestCreating
 	}
 
-	if loadTestStatus.Phase == loadTestV1.LoadTestErrored ||
-		loadTestStatus.Phase == loadTestV1.LoadTestFinished {
```
s4nji, line 210: I don't think this should be removed; once loadtest is marked as Finished, it's over and its status should not be updated again.

I believe the addition in L242 already fixes the bug where the loadtest is not being deleted.

s-radyuk: Since `Job Status` is not persisting on the LoadTest object, it is not saved after we finish processing this LoadTest object in syncHandler. L242 updates the status but it turns to `nill` as soon as we finish processing this object. And the next time controller picks this object, it will drop it from SyncStatus as finished, and `checkLoadTestLifeTimeExceeded` will drop it too, because `loadTest.Status.JobStatus.CompletionTime == nil`. And that's how this object gets stuck in the cluster forever.  
Screenshot 2021-06-02 at 17 36 23

diegomarangoni (approved):
lgtm

pkg/backends/locust/backend.go
s4nji: Looks like we need to find out why `LoadTest.Status.JobStatus` is nil on every `syncHandler` iteration, but that is a separate issue.

s4nji (approved):

PR #141, opened 2021-06-04
starting a local server and sending requests to it from a real test

s-radyuk:
This PR updates the integration tests.
- Starting a local server in kind
- Adding a real JMeter test to send load to this server
- Checking how many requests were actually sent


s4nji (approved):
👍 lgtm

PR #147, opened 2021-06-11
Feature/issue 146 image in url

jdani:
Now it is possible to define the image to use for the master and worker pods.

This PR includes:

- Code
- Documentation updated
- openapi.json updated


s-radyuk:
Hey @jdani ,
Thank you for your contribution!
Can you please resolve the conflicts? And maybe add a test to check that the image provided in the request is used for the created loadtest, not the image set as env var on the controller level?

docs/user-flow.md
```diff
@@ -49,6 +49,24 @@ curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
...
+curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
+  -H 'Content-Type: multipart/form-data' \
+  -F distributedPods=1 \
+  -F testFile=@examples/constant_load.jmx \
+  -F testData=@artifacts/loadtests/testData.csv \
+  -F envVars=@artifacts/loadtests/envVars.csv \
+  -F type=JMeter \
+  -F tags=tag1:value1,tag2:value2 \
+  -F overwrite=true \
+  -F masterImage=masterContainerImage:latest \
```
s-radyuk, line 64: Since we use this documentation as a user guide, we can add a real image to this example, e.g. `hellofresh/kangal-jmeter-worker:5.4.1` and `hellofresh/kangal-jmeter-master:5.4.1`

jdani:
Done!

jdani:
@s-radyuk  Hi! Not rushing at all, but I would like to be sure the PR is ready for review. Last time I thought it was but I was wrong.

Thanks!

pkg/proxy/request_test.go
```diff
@@ -429,3 +429,96 @@ func TestGetDuration(t *testing.T) {
...
+
+			actualImage := image.Image + ":" + image.Tag
+
+			assert.Nil(t, err)
+
+			if ti.expectError {
+				// assert.Error(t, err)
+				assert.NotEqual(t, expectedImage, actualImage)
+			} else {
+				// assert.NoError(t, err)
```
s-radyuk, line 518: Is it a leftover?

lucasmdrs:
Hi @jdani, thank you for taking the time to do this contribution! :heart: 
I'll take a look at it right away :eyes:

pkg/proxy/request.go
```diff
@@ -268,6 +286,25 @@ func getDuration(r *http.Request) (time.Duration, error) {
 	return time.ParseDuration(val)
 }
 
+func getImage(r *http.Request, role string) (apisLoadTestV1.ImageDetails, error) {
```
lucasmdrs, line 289: maybe you could remove the error return here, as it returns a empty string by default

pkg/proxy/request_test.go
jdani: No. There are two checks there:

- Line 512: Assert getImage() returns no error

- From Line 512 - 520: Check that the value returned by getImage() is equal to the expected value or not.

pkg/proxy/request_test.go
```diff
@@ -429,3 +429,96 @@ func TestGetDuration(t *testing.T) {
...
+		t.Run(ti.tag, func(t *testing.T) {
+
+			image := apisLoadTestV1.ImageDetails{
+				Image: "",
+				Tag:   "",
+			}
+
+			err := error(nil)
+
+			sentImage := ti.imageName + ":" + ti.imageTag
```
lucasmdrs, line 498: You could provide the complete image name in this test scenarios, it would help to check cases when the `strings.Contains(imageStr, ":")` is false and the default value should be used.

lucasmdrs (commented):
Left some comments on a couple of things but overall looks good to me. :+1: 
Please reply on them if you agree or not, you can also @ me anytime.

pkg/proxy/request.go
jdani: Done!

pkg/proxy/request_test.go
lucasmdrs: Sorry, didn't mean to refer to the default value from the backend, just the default returned from `getImage` when `strings.Contains(imageStr, ":")` is false, which should be an empty `apisLoadTestV1.ImageDetails` struct

lucasmdrs:
@jdani I would also ask you to create a feature flag for this if possible.
This changes will enable any image to be used by any user that are able to do a HTTP request for Kangal Proxy, which might create a security issue. If we ensure this functionality is disabled by default we can work on something else to tackle this problem afterwards

jdani:
It makes sense. Help me a bit with this, please. What do you mean exactly with feature flag? Is there any feature flag already implemented?

Will it work an environment variable? This way it can be configured easily via helm values too.

Thanks

pkg/proxy/request_test.go
jdani: Ok ok, I get it. I will add a condition on the test so if any of image name or image tag is empty, an empty `apisLoadTestV1.ImageDetails` will be returned

lucasmdrs: a good condition to test would be providing at the request `someImage/withNoColon` this should also output an empty `apisLoadTestV1.ImageDetails` as `strings.Contains(imageStr, ":")` would be false :wink:

lucasmdrs:
> It makes sense. Help me a bit with this, please. What do you mean exactly with feature flag? Is there any feature flag already implemented?

I mean feature toggle* actually, there's no implementation of this yet, but you could use `proxy.Config` and set this up as an `envconfig` field

> Will it work an environment variable? This way it can be configured easily via helm values too.
Yes, exactly ! :smile:

pkg/proxy/request_test.go
lucasmdrs: I believe it was referring to the commented assertions. But this should not be necessary anymore, as there's no error output from the `getImage` function after the last changes

pkg/proxy/request.go
lucasmdrs: There are some `if err != nil ` checks after the function call, I think those can be removed as well

pkg/proxy/request_test.go
jdani: Completely right. In fact it's removed in the last commit

jdani: Fixed!

pkg/proxy/request.go
jdani: Fixed

pkg/proxy/request_test.go
jdani: Yep, and it did not work like that. Empty apisLoadTestV1.ImageDetails is the expected return for empty or not valid images.
I was not very happy with the way the image url string was being treated, so now I use the module `github.com/novln/docker-parser`. It is included in the vendor and go.mod.

The suggested test is implemented as well.

jdani:
> I mean feature toggle* actually, there's no implementation of this yet, but you could use `proxy.Config` and set this up as an `envconfig` field

I expect to code this in the backend, not in the proxy. Something as simple as checking in `fromHTTPRequestToLoadTestSpec` if an envconfig exists. If so, call getImage. Otherwise, masterImage (`mi`) and workerImage (`wi`) will be `""`, so the default image will be used. Won't it work?

pkg/proxy/request_test.go
lucasmdrs: I don't see this lib in the modules :thinking: 
If I may advise you, don't worry too much about this and go for the simpler approach.. 

I think you were doing great by enforcing a `imagename:tag` pattern, it's the most common representation of an image so it's a very reasonable decision, as long we're clearly informing the user what we expect as an input (which you are already doing it with the regex in the validation) :relieved:

lucasmdrs:
> OK! That makes sense to me. Let me work on it...

Awesome! thank you very much :smile:  

> @lucasmdrs I need some guidance on this. The thing is that the cfg object is not available on the `proxy/request.go` file. What would you think about accessing the env var using os.Getenv()? Otherwise, what is your suggestion? How to access envconfig fields within the request file?

Indeed, but the `fromHTTPRequestToLoadTestSpec` function is private and only get's invoked by `*Proxy.Create`, I think we could add this `bool` attribute to the Proxy structure and just pass it as an extra arg for the function, something like:
```golang
...
	// Making valid LoadTestSpec based on HTTP request
	ltSpec, err := fromHTTPRequestToLoadTestSpec(r, logger, p.customImageEnabled)
	if err != nil {
...
```
or a better attribute name :laughing: 

At the time of the creation of the Proxy struct (`NewProxy`), you have the cfg in hand so that would be a way of "cascading" this information :wink: What do you think ?

jdani:
Done as suggested. And it is working, I've tested it. But I think it is still not ready to go. There are no tests for the feature flag.

I've tried in `pkg/proxy/proxy_test.go`, using as a base any of the tests there, but once the proxy is created and the request sent, the response does not includes any info related to custom images, so not sure how to test it. I thought about modifying the request to include that info, as it would be useful, but this is taking quite longer than expected...

jdani:
OK, now there are tests even for the custom image feature flag. Hope this seems fine to you. For me it looks like finished :P

s-radyuk (approved):
Thanks a lot for your contribution @jdani !

lucasmdrs (approved):
:tada:

PR #150, opened 2021-06-30
Ghz toml config files

lucasmdrs:
Add support to toml configuration files config for ghz backend


pkg/backends/ghz/backend.go
```diff
@@ -133,7 +133,8 @@ func (b *Backend) Sync(ctx context.Context, loadTest loadTestV1.LoadTest, report
 	)
 
 	// Create testfile ConfigMap
-	tfCfgMap, err := NewFileConfigMap(loadTestFileConfigMapName, configFileName, loadTest.Spec.TestFile)
+	configFile := fmt.Sprintf("%s.%s", configFileName, guessTypeFromContent(loadTest.Spec.TestFile))
```
s4nji, line 136: I think this should be `configFileName`, since it really is just the filename, not the file itself.  
We can remove the const and rename this to `configFileName`.

pkg/backends/ghz/backend.go
```diff
@@ -168,16 +169,23 @@ func (b *Backend) Sync(ctx context.Context, loadTest loadTestV1.LoadTest, report
...
 		mounts = append(mounts, m)
 	}
 
+	// Build job args
+	args, err := NewCommandArgs(configFile)
+	if err != nil {
+		b.logger.Error("Error creating job args", zap.String("configFile", configFile), zap.Error(err))
+		return err
+	}
+
```
s4nji, line 186: Since this is only used in `NewJob()` and is not a kube resource like the things above; what do you think about moving this as a private function in `resources.go` and calling that in `NewJob()` directly?

We can guess the config extension in it as well.

lucasmdrs: > We can guess the config extension in it as well.

As it's being referenced in both configmap and volume creation, I kept it at the `backend.go`, do you see a better way @s4nji ?

s4nji: Looks like we can just drop the extension for the file name: it seems to [try to parse][1] it in toml and json if it has no matching extension 🤔 

[1]: https://github.com/jinzhu/configor/blob/591dd1903f49b827983691a64ec1e40ff53d05bc/utils.go#L101

s4nji: We can leave a comment somewhere noting that ghz will try to guess the config file format for posterity

s4nji (approved):
lgtm

PR #158, opened 2021-09-13
EES-6536 - replace unverified lint actions with super linter

s-radyuk:
This PR updates the testing workflow with verified actions.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -18,8 +18,6 @@ builds:
 dockers:
   - goos: linux
     goarch: amd64
-    binaries:
-      - kangal
     image_templates:
       - "hellofresh/kangal:latest"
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -18,8 +18,6 @@ builds:
 dockers:
   - goos: linux
     goarch: amd64
-    binaries:
```
lucasmdrs, line 21: [the docs](https://goreleaser.com/deprecations/#dockerbinaries) say to replace `binaries` with `ids`, not sure if removal will not cause an issue

s-radyuk: I tried to replace it first, as the docs say, but it caused another issue. Removal worked for me both locally and here.

lucasmdrs (approved):

s4nji (approved):

PR #169, opened 2021-12-27
EES-4738 - Document all the environment variables used in Kangal

s-radyuk:
This PR adds a new document which contains all the env vars used to configure Kangal.
Documentation for helm chart was also updated.
Some unused and duplicated variables were removed from configs.
Also superlinter version was updated in the testing workflow.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,7 +32,7 @@ jobs:
           VALIDATE_GITHUB_ACTIONS: false
           VALIDATE_NATURAL_LANGUAGE: false
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          KUBERNETES_KUBEVAL_OPTIONS: "--ignore-missing-schemas"
+          VALIDATE_KUBERNETES_KUBEVAL: false
 
       - name: Get tag
         id: get_tag
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,6 +32,7 @@ jobs:
           VALIDATE_GITHUB_ACTIONS: false
           VALIDATE_NATURAL_LANGUAGE: false
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          KUBERNETES_KUBEVAL_OPTIONS: "--ignore-missing-schemas"
 
       - name: Get tag
         id: get_tag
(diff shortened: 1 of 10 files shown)


charts/kangal/README.md
```diff
@@ -9,20 +9,20 @@ This chart bootstraps a [kangal](https://github.com/hellofresh/kangal) deploymen
...
 ```shell
-$ helm install kangal kangal/kangal
+ helm install kangal kangal/kangal
 ```
 
 > **Tip**: You can provide values on the command line with `--set` or using the `values` file with `-f my-values-file.yaml`. Eg.:
 To set AWS credentials using command line you can use the following flags:
 
-```
+```yaml
```
vgarvardt, line 25: this is not yaml, I think, looks like shell command flags

.github/workflows/testing.yml
```diff
@@ -24,14 +24,15 @@ jobs:
           version: v1.42.1
 
       - name: Run General Linting
-        uses: docker://ghcr.io/github/super-linter:slim-v4
+        uses: docker://ghcr.io/github/super-linter:slim-v4.8.5
```
vgarvardt, line 27: in theory this should be the same images - https://github.com/github/super-linter/pkgs/container/super-linter
image

But I'm fine with pinning the full version as it can be some sort of caching preventing it from using really this latest version

.github/workflows/testing.yml
```diff
@@ -24,14 +24,15 @@ jobs:
...
         env:
           VALIDATE_ALL_CODEBASE: false
           VALIDATE_GO: false
           VALIDATE_JSCPD: false
           VALIDATE_GITHUB_ACTIONS: false
           VALIDATE_NATURAL_LANGUAGE: false
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VALIDATE_KUBERNETES_KUBEVAL: false
```
vgarvardt, line 35: Probably we should switch from opt-out model to opt-in and list all the things we want to lint explicitly. Btw - why do we have `VALIDATE_GO: false` if this is the main project language? =) Anyway, should be done in own PR

vgarvardt (approved):

.github/workflows/testing.yml
s-radyuk: `VALIDATE_GO: false` is because we use golangci-lint-action for linting go. And after that we use superlinter for all the rest.

vgarvardt: `super-linter` uses `golangci`under the hood, so it makes sense to use the super one

PR #172, opened 2022-01-03
Clean up request validations

s-radyuk:
Current problem: When POST request is sent, kangal first runs httpValidator method which uses govalidator. But after this we validate all the fields one more time with custom validation methods. 

This PR:
- removes govalidator in order to avoid duplicated validations and reduce dependencies
- unifies all the error messages for all the request fields
- updates tests
- moves some methods and vars from `proxy.go` to `request.go`, where they are actually used


pkg/proxy/request.go
```diff
@@ -200,17 +205,26 @@ func getEnvVars(r *http.Request) (map[string]string, error) {
...
-			return "", err
+	for _, f := range testDataFileFormats {
+		if fileType == f {
+			if fileType == "csv" {
+				err = checkCsvFile(stringTestData)
+				if err != nil {
+					return "", err
+				}
+			}
+			return stringTestData, nil
```
lucasmdrs, line 223: You could change `testDataFileFormats` to be a map of format: checker (`map[string]func(string) (string,error)`) that way the loop and nested ifs are avoided, looking something like this in the end:

```golang
if formatCheck, acceptedFormat := testDataFileFormats[fileType]; acceptedFormat {
    return formatCheck(stringTestData)
}
return "", ErrWrongFileFormat
```

and the map being:
```golang
type formatCheckFunc func (string) (string,error)

var testDataFileFormats = map[string]formatCheckFunc{
    "csv" : func (s string) (string, error) {
        if err = checkCsvFile(stringTestData); err != nil {
            return "", err
        }
        return s, nil
    },
    "protoset": func (s string) (string,error) { return s, nil }
}

File: pkg/proxy/request.go [hunk 2]
```diff
@@ -228,18 +242,22 @@ func checkCsvFile(s string) error {
...
+	content, fileType, err := getFileFromHTTP(r, testFile)
+	if err != nil {
+		return "", err
+	}
+
+	for _, f := range testFileFormats {
+		if fileType == f {
+			return content, err
+		}
+	}
```
lucasmdrs, line 254: this could be a map[string]bool too, to avoid the loop/if :
```golang
var testFileFormats = map[string]bool{
    "jmx":  true,
    "py":   true,
    "json": true,
    "toml": true
}
```


```golang
if accepted := testFileFormats[fileType]; !accepted {
    return "", ErrWrongFileFormat
}
return content, nil
```

pkg/proxy/request.go
```diff
@@ -112,49 +109,49 @@ func fromHTTPRequestToLoadTestSpec(r *http.Request, logger *zap.Logger, allowedC
 	o, err := getOverwrite(r)
 	if err != nil {
 		logger.Debug("Bad value: ", zap.String("field", overwrite), zap.Bool("value", o), zap.Error(err))
-		return apisLoadTestV1.LoadTestSpec{}, fmt.Errorf("bad %q value: should be bool", overwrite)
+		return apisLoadTestV1.LoadTestSpec{}, fmt.Errorf("bad %s value: should be bool", overwrite)
```
lucasmdrs, line 112: I personally like the quotes from `%q`, I think it makes more clear that the field name is not part of the string.. but the message could be improved too, something like:
 ```golang
 fmt.Errorf("field %q should have a boolean value", overwrite)
 ```

pkg/proxy/request.go
```diff
@@ -330,18 +316,22 @@ func getDuration(r *http.Request) (time.Duration, error) {
 	return time.ParseDuration(val)
 }
 
-func getImage(r *http.Request, role string) apisLoadTestV1.ImageDetails {
-
+func getImage(r *http.Request, role string) (apisLoadTestV1.ImageDetails, error) {
 	imageStr := r.FormValue(role)
 
+	imageRegex := regexp.MustCompile("^.*:.*$|^$")
```
lucasmdrs, line 322: could you move this to a package `var` like the others in line 27 ?
the regexp compiling is a resource expensive process but it need to be done only once.

pkg/proxy/request.go
s-radyuk: Following your suggestion below, I prefer to represent `testDataFileFormats` as map, the same way as `testFileFormats`, but not to add a `formatCheckFunc` inside the map. I think suggested code is a bit more difficult to read than a separate check applied only to csv files.

pkg/proxy/request.go
s-radyuk: I didn't like the slashes in the error message like this:
`
"error": "bad \"distributedPods\" value: should be integer"
`

pkg/proxy/request.go
s-radyuk: Thanks for noticing it! Moved to vars.

lucasmdrs (approved):

vgarvardt (approved):

Issue #139, opened 2021-06-01
Larger test data files


thiagoarrais:
When trying to start a test with a ~1M test data (CSV) file, I've noticed that Kangal silently rejects the test. There is an error message printed to the controller logs, but the POST request yields a 201 response and the error can only be externally detected by examining the response to a succeeding GET to `/load-test/:test_name` and noticing that it does not include a test name.

Besides warning that some kind of error message should be returned to the end user, I'd like to start a discussion here: do we see that test data size should be limited? It can be argued that we don't need a lot of test data since most load testing can be done by just repeating requests. On the other hand, there are very legitimate use cases for large data sets in a load test scenario like cache busting.

diegomarangoni:
Hey @thiagoarrais, thanks for reporting.

Kangal stores test data on a Kubernetes ConfigMap and there is a 1MB limit on those resources.

> A ConfigMap is not designed to hold large chunks of data. The data stored in a ConfigMap cannot exceed 1 MiB. If you need to store settings that are larger than this limit, you may want to consider mounting a volume or use a separate database or file service.

https://kubernetes.io/docs/concepts/configuration/configmap/

So there is not much to be done unless change the whole approach to use Volumes instead.
But we could improve the experience a bit and fail fast in case of the file exceeds the size.

thiagoarrais:
I see. Would kangal be willing to accept a PR that moves the data to Volumes then?

diegomarangoni:
Definitely @thiagoarrais,
If you prefer, you can propose the solution first and then we can discuss it.

I think the main question is, where those files will be stored?
Proxy already has S3 integration, so we could rely on this maybe.

amandahla:
Another idea that I'm working on: convert the csv to a gzip 64 file and convert back using an init container.

PR #177, opened 2022-01-20
Allow large test data files (#139)

amandahla:
- Convert CSV to GZIP BASE64 format
- Create InitContainer that convert back to CSV and copy to volume so worker can access the CSV


s-radyuk:
Hello @amandahla ! Thank you very much for your interest to Kangal!
I would like to ask you to add an integration test to your PR to make sure your change helps with large testdata files.

amandahla:
@s-radyuk Great project :-) 
I'm not sure if it's ok but instead of create a new integration test, I have just created a new testdata.csv file named "testdata-jmeter.csv" (1332K).  WDYT?

s-radyuk:
Hello again @amandahla Sorry for the delay with the answer. 

Thank you for paying attention to the problem with large testdata files. We also investigated it before and found some limitations on the Kubernetes side. Here are some cases that we’ve tried and their errors:

1. Create a loadtest with testdata file with size > 3Mb. The error thrown will be the following:
`"error": "Request entity too large: limit is 3145728"`. 
It's coming from Kubernetes API preventing me from creating a very big kubernetes object. 

2. Create a loadtest with testdata file 2Mb < size < 3Mb. The error thrown will be the following:
`"error": "rpc error: code = ResourceExhausted desc = trying to send message larger than max (2493228 vs. 2097152)"`

3. Create a loadtest with testdata file size < 2097152.  The error thrown will be the following: 
`"error": "etcdserver: request is too large"`

I saw your [proposal](https://github.com/hellofresh/kangal/issues/179) and I totally agree that dynamic volume provisioning with ReadWriteMany can help us in this case. 

I also had a chance to try out the code from this PR and I've noticed the following change in behaviour (unfortunately our integration tests don't show this problem):

Config map with testdata now contains testdata.csv.gz, which is expected with the change from this PR. But for some reason the data is not extracted to the `/testdata` directory. Because of it JMeter process running in worker pod throws the following error in logs: 
`Test failed! MDC{}
java.lang.IllegalArgumentException: Could not read file header line for file /testdata/testdata.csv`
The JMeter process expects to find the data in CSV format in `/testdata/testdata.csv` but this directory is empty. This prevents a loadtest from running as designed.

amandahla:
Hi @s-radyuk , thanks for testing!

I'm really sorry but I could swear that I tested this code and it worked fine (maybe it's too hot here in Brazil and it's affecting my brain). Anyway, I have changed a few things and now I suppose that should work with bigger CSV files.

I had to modify the Spec and I'm not sure if I did it in the right place.

To work with ingress-nginx (for example), it would be necessary to change "proxy-body-size" via annotations or configmap like explained [here](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size)
I did it using this command:
```bash
kubectl annotate --overwrite ingress kangal-proxy nginx.ingress.kubernetes.io/proxy-body-size="10M"
```


```bash
kubectl annotate --overwrite ingress nginx.ingress.kubernetes.io/proxy-body-size=10M --all
```

pkg/backends/jmeter/backend.go
```diff
@@ -154,6 +157,19 @@ func (b *Backend) TransformLoadTestSpec(spec *loadTestV1.LoadTestSpec) error {
...
+		var by bytes.Buffer
+		gz := gzip.NewWriter(&by)
+		if _, err := gz.Write([]byte(spec.TestData)); err != nil {
+			return err
+		}
+		if err := gz.Close(); err != nil {
+			return err
+		}
+		spec.TestData = base64.RawStdEncoding.EncodeToString(by.Bytes())
+	}
```
lucasmdrs, line 171: would be great to extract this :+1:

pkg/proxy/proxy.go
```diff
@@ -181,6 +181,11 @@ func (p *Proxy) Create(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// TODO review if this is the right way to do it
```
lucasmdrs, line 184: I believe this should be addressed by the [BuildLoadTestObject](https://github.com/hellofresh/kangal/blob/c120604e26f376f132bcd0e3ec29e22c994a1f77/pkg/proxy/proxy.go#L123) method, keeping the Proxy free of specific backend implementations

lucasmdrs (commented):
Hi @amandahla :wave: 

Thank you for taking the time for this contribution so far, we really appreciate it!

I left some small comments/suggestions, if you address them and fix the conflicts I'll make sure to review it, to move this PR forward :slightly_smiling_face:

pkg/proxy/proxy.go
amandahla: @thiagoarrais found out that if we change the TransformLoadTestSpec to the top, that would make this "if" useless.

thiagoarrais: If BuildLoadTestObject did the compressing itself, that would also be unnecessary. But I believe that would affect every backend instead of just JMeter. Maybe we want that?

lucasmdrs (approved):
Thanks for fixing the conflicts, just tested all locally and it seems to be working nicely. :heavy_check_mark: 

Thank you very much your contribution ! 
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

Issue #196, opened 2022-02-16
Move Helm chart to separate repository


lucasmdrs:
Keeping both Kangal releases and helm charts releases in the same repository caused problems with goreleaser #170  and confusion on users #182 

A proposal for this issue would be to move the [helm chart to separate repository](https://artifacthub.io/packages/helm/kangal/kangal) and leave Kangal repository just with the source code.

amandahla:
Hi @lucasmdrs 
Will new versions still be published on docker hub? We would like to upgrade our environment to support custom data.

lucasmdrs:
Hi @amandahla yes, the idea is just to extract the things related to Chart to another repository. Docker image would still be published to the same place and the versions would continue the same semantic versioning number

PR #206, opened 2022-02-24
Update docs warning about csv data split between distributed pods

amandahla:
Avoid confusion about test data usage and clarify that won't be duplicated data.


lucasmdrs:
I'll take a look at this Lint issue from Go Releaser, it's kind of a :chicken: :egg: problem with the releases.. related to #196

PR #209, opened 2022-04-14
Use BinaryData on ConfigMap to allow binary data for tests source

walterwanderley:
It's needed to support [k6 archive](https://k6.io/docs/misc/archive-command/) files.


amandahla:
@lucasmdrs friendly ping :-)

lucasmdrs:
Hi @walterwanderley @amandahla  :wave:

Thanks for the contribution! :blue_heart:   ( and the friendly ping :sweat_smile: )

Apologies for the delayed reply, this changes would affect the CRD and I would like to discuss a little bit with our team before proceeding, please bear with us a little longer :pray:

We'll get back to you as soon as we can :runner:

walterwanderley:
Hi @lucasmdrs!
Could you guide me how to generate the v2 of the LoadSpec?

lucasmdrs:
> Hi @lucasmdrs! Could you guide me how to generate the v2 of the LoadSpec?

The codegen for a new CRD version consists in three steps:

1. Creating a new directory in the `pkg/kubernetes/apis/loadtest` with the version name e.g. `v2`
2. Adding the `docs.go` `register.go` and `types.go` with the definitions similar to what we have in `v1` but with the respective changes.
3. Update the `hack/update-codegen.sh` script to include the new types e.g `loadtest:v2` and run `make update-codegen`

However, after some thought on the subject I don't think we want backward compatibility (make it work with both v1 and v2). Because our backends and code structure was not designed to support this. I'm considering merging the PR as it is and try notify the best possible about compatibility issues when releasing the new version

lucasmdrs (approved):
@walterwanderley We are thankful for your patience with the time it took us with this change and appreciate a lot your contribution. Let's :shipit: !

Issue #210, opened 2022-04-18
Avoid putting worker pod in the same node


amandahla:
We have noticed that all pods were created at same node in the cluster.

Suggestion:
Add podAntiAffinity at spec.

[Reference here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#more-practical-use-cases)

PR #211, opened 2022-04-18
Add Affinity Spec for K6 and JMeter #210

amandahla:
Problem:
Pods are being created at the same node, overloading it

Proposal:
Define Affinity Spec at Pod

- Where would be better to define that? Like proxy and openapi affinity configuration
- We have found that all pods are created with same label "app=loadtest-worker-pod". It's not possible to use label-selector to create a pod in a node with no pods from the same test.

### DIFF
--- a/pkg/backends/jmeter/backend.go
+++ b/pkg/backends/jmeter/backend.go
@@ -310,6 +310,10 @@ func generateBase64(testData string) (string, error) {
 		return result, err
 	}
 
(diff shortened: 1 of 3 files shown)


lucasmdrs:
That's a really good suggestion :smiley: 

resolves #210

lucasmdrs (approved):
Thank you for the contribution :green_heart:

Issue #214, opened 2022-05-16
Fix kubernetes code gen 


lucasmdrs:
Currently the code generator script is failing, preventing new contributions to make changes in the CRD.

amandahla:
Maybe update-codegen.sh shouldn't do "rm -rf "$TEMP"" at the end?

lucasmdrs:
Hi @amandahla :wave: 

Thank you for taking the time to check this, that would allow the verification script to work again :+1: 

But I think we'll need to go a bit deeper and improve this process. We don't have much documentation about it and we could automate a few things, for example, the full procedure is:

 - Vendor dependencies `go mod vendor`
 - Grant executable permission to code-generator `chmod +x vendor/k8s.io/code-generator/*.sh`
 - Run the update/verification `./hack/update-codegen.sh`and `./hack/verify-codegen.sh`
 
Probably reducing a few steps and adding this to the Makefile could be a better approach. :robot: 

Also we need to add a note about backward compatibility in the docs, changes to generated-code would likely represent changes to CRDs and a brake all previous installations :memo:

amandahla:
Totally agree @lucasmdrs 
Can I help with this issue?
- [ ] Update docs
- [ ] Automate procedure using Makefile

lucasmdrs:
Yes! Any contribution on this topic is much appreciated :pray:

lucasmdrs:
@amandahla I'll be working on this next week
I'll review your changes and add some checks to our workflows for this too.

Hopefully that will unblock some of the PRs we have that change the CRD like  #225 and #209

PR #215, opened 2022-05-24
Add verify-codegen to Makefile and update doc (#214)

amandahla:
update-codegen.sh was removing TMP dir at the end causing verify-codegen.sh to fail while comparing results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Folders
 _obj
 _test
+_tmp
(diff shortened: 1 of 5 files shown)


lucasmdrs (dismissed):

lucasmdrs:
Hi @amandahla 
Thank you so much for the contribution, I have included the commit in the #230 merge

PR #222, opened 2022-06-29
Take into account worker and master images for Locust

st3veV:
Use master and worker images for Locust runner if specified.

This is very useful if using Locust Master and Locust Boomer workers.

### DIFF
--- a/pkg/backends/locust/resources.go
+++ b/pkg/backends/locust/resources.go
@@ -81,9 +81,9 @@ func newMasterJob(
 
 	ownerRef := metaV1.NewControllerRef(&loadTest, loadTestV1.SchemeGroupVersion.WithKind("LoadTest"))
 
-	imageRef := fmt.Sprintf("%s:%s", loadTest.Spec.MasterConfig.Image, loadTest.Spec.MasterConfig.Tag)
+	imageRef := fmt.Sprintf("%s:%s", image.Image, image.Tag)
 	if imageRef == ":" {
-		imageRef = fmt.Sprintf("%s:%s", image.Image, image.Tag)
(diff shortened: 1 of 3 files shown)


pkg/backends/locust/resources.go
```diff
@@ -227,9 +227,9 @@ func newWorkerJob(
 
 	ownerRef := metaV1.NewControllerRef(&loadTest, loadTestV1.SchemeGroupVersion.WithKind("LoadTest"))
 
-	imageRef := fmt.Sprintf("%s:%s", loadTest.Spec.MasterConfig.Image, loadTest.Spec.MasterConfig.Tag)
+	imageRef := fmt.Sprintf("%s:%s", image.Image, image.Tag)
 	if imageRef == ":" {
-		imageRef = fmt.Sprintf("%s:%s", image.Image, image.Tag)
+		imageRef = fmt.Sprintf("%s:%s", loadTest.Spec.MasterConfig.Image, loadTest.Spec.MasterConfig.Tag)
```
lucasmdrs, line 232: should it be the Worker image instead ?

lucasmdrs (commented):
Hey @st3veV :wave: 

Thank you for  the contribution! I left a small comment :speech_balloon:

pkg/backends/locust/resources.go
st3veV: Hey @lucasmdrs,

hmm, I only flipped the conditions to look at the supplied `image` argument first and then fallback to Master image to keep the behavior backwards compatible. Following the logic "if the worker is not specified use the master".

Now the question is basically what behavior is desired - the one I mentioned above or "if worker is not specified use the default"?

(sorry for late response, I was on vacation)

lucasmdrs:
Thanks again for your contribution @st3veV :sparkles:

lucasmdrs (approved):

PR #224, opened 2022-07-25
Use docker reference for image parsing

st3veV:
Previous parsing wasn't working with deeply nested repositories and only supported `"://:"`.
Instead of making this parsing more complex I replaced it with using standard docker image reference which should be bullet-proof enough for all the edge cases of custom repositories and image tags.
Changes in tests reflect this as well because the tests were using `masterImage` as a name but only lowercase is allowed in the official spec.


lucasmdrs:
Hey @st3veV :wave: 
Thanks for the PR, however this library seems to be extremely vulnerable based on the [WhiteSource Security Check](https://github.com/hellofresh/kangal/pull/224/checks?check_run_id=7499952589) I think it would be best to wait for a [release that address those issues](https://github.com/distribution/distribution/issues/3677) before including it

st3veV:
Hey @lucasmdrs,
that is completely fair, I totally didn't expect the official distribution library to be this riddled with CVEs. The vulnerabilities seems mostly related to yaml parsing and encryption and neither of them is actually used but your concern is understandable. I'm fine using a fork for now because I'll be using kangal in a closed cluster without being open to the internet.

I have an alternative solution to the parsing trouble in mind and that is to remove parsing of the `image:tag` altogether. The reason is that it doesn't have any other purpose than request validation and fallback to defaults. Fallback to defaults could be solved by simply checking for `""` and I could imagine skipping the validation for this because if there is some invalid value k8s will complain anyway (and some invalid values were already allowed to get through without it being an issue). What do you think?

lucasmdrs:
Given that we have a flag to prevent/allow custom images and the current validation logic is not accepting valid image values, I agree with removing the parsing :+1:

st3veV:
Alright, see https://github.com/hellofresh/kangal/pull/225
Closing this one.

PR #240, opened 2022-12-01
Removing unused metrics for loadtests counting

s-radyuk:
This PR removes unused metric to count started and finished load tests. These metrics were replaced in #233 by MRunningLoadtestCountStat used to count currently running load tests.


lucasmdrs (commented):
maybe we should modify the `MRunningLoadtestCountStat` to be more generic, like `MLoadtestCountStat` and report the phase through `tag` (labels)

s-radyuk:
@lucasmdrs That's exactly what I want to do in a different PR by separate task.

PR #250, opened 2022-12-21
extending lt counter metric

s-radyuk:
This PR extends the logic of existing CountRunningLoadtests to count all the tests currently existing in the cluster:
- rename metric to CountExistingLoadtests
- metric returns count of loadtests by type and by phase
- the same method CountExistingLoadtests now can be used instead of CountActiveLoadTests with duplicated logic

### DIFF
--- a/pkg/kubernetes/client.go
+++ b/pkg/kubernetes/client.go
@@ -232,33 +232,33 @@ func BuildClientConfig(masterURL string, kubeConfigPath string, timeout time.Dur
 }
 
 // CountExistingLoadtests used in metrics to report running loadtests
-func (c *Client) CountExistingLoadtests() (map[string]int64, map[string]int64, error) {
+func (c *Client) CountExistingLoadtests() (map[apisLoadTestV1.LoadTestPhase]int64, map[apisLoadTestV1.LoadTestType]int64, error) {
 	tt, err := c.ltClient.List(context.Background(), metaV1.ListOptions{})
 	if err != nil {
 		c.logger.Error("Couldn't list existing loadtests", zap.Error(err))
 		return nil, nil, err
 	}
 
-	var phaseCount = map[string]int64{
-		apisLoadTestV1.LoadTestRunning.String():  0,
-		apisLoadTestV1.LoadTestFinished.String(): 0,
-		apisLoadTestV1.LoadTestCreating.String(): 0,
-		apisLoadTestV1.LoadTestErrored.String():  0,
-		apisLoadTestV1.LoadTestStarting.String(): 0,
+	var phaseCount = map[apisLoadTestV1.LoadTestPhase]int64{
+		apisLoadTestV1.LoadTestRunning:  0,
+		apisLoadTestV1.LoadTestFinished: 0,
+		apisLoadTestV1.LoadTestCreating: 0,
+		apisLoadTestV1.LoadTestErrored:  0,
+		apisLoadTestV1.LoadTestStarting: 0,
 	}
 
-	var typeCount = map[string]int64{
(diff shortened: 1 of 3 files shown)


pkg/kubernetes/client.go
```diff
@@ -248,20 +231,59 @@ func BuildClientConfig(masterURL string, kubeConfigPath string, timeout time.Dur
...
+
+	var phaseCount = map[string]int64{
+		"running":  0,
+		"finished": 0,
+		"creating": 0,
+		"errored":  0,
+		"starting": 0,
+	}
+
+	var typeCount = map[string]int64{
```
fabioserra, line 250: Use LoadTestType here instead of string and the constants instead of the key values.
"k6" -> LoadTestTypeK6

pkg/kubernetes/client.go
```diff
@@ -248,20 +231,59 @@ func BuildClientConfig(masterURL string, kubeConfigPath string, timeout time.Dur
...
+		}
+		if loadTest.Status.Phase.String() == "creating" {
+			phaseCount["creating"]++
+		}
+		if loadTest.Status.Phase.String() == "finished" {
+			phaseCount["finished"]++
+		}
+
+		if loadTest.Spec.Type == apisLoadTestV1.LoadTestTypeK6 {
+			typeCount["k6"]++
```
fabioserra, line 275: Use the constant as key value in the typeCounters

pkg/kubernetes/client.go
```diff
@@ -248,20 +231,59 @@ func BuildClientConfig(masterURL string, kubeConfigPath string, timeout time.Dur
...
+// CountExistingLoadtests used in metrics to report running loadtests
+func (c *Client) CountExistingLoadtests() (map[string]int64, map[string]int64, error) {
 	tt, err := c.ltClient.List(context.Background(), metaV1.ListOptions{})
 	if err != nil {
 		c.logger.Error("Couldn't list existing loadtests", zap.Error(err))
-		return 0
+		return nil, nil, err
+	}
+
+	var phaseCount = map[string]int64{
```
fabioserra, line 242: Use LoadTestPhase here instead of string and the constants instead of the key values.
"creating" -> LoadTestCreating

pkg/kubernetes/client.go
```diff
@@ -248,20 +231,59 @@ func BuildClientConfig(masterURL string, kubeConfigPath string, timeout time.Dur
...
+		"locust": 0,
+		"ghz":    0,
 	}
 
-	var rt = 0
 	for _, loadTest := range tt.Items {
-		if loadTest.Status.Phase == apisLoadTestV1.LoadTestRunning {
-			rt++
+		if loadTest.Status.Phase.String() == "running" {
+			phaseCount["running"]++
```
fabioserra, line 259: Use the constant as key value in the phase counter

fabioserra (changes requested):

pkg/kubernetes/client_test.go
```diff
@@ -384,9 +382,10 @@ func TestCountActiveLoadTests(t *testing.T) {
 
 	logger := zap.NewNop()
 	c := NewClient(loadtestClientset.KangalV1().LoadTests(), kubeClientSet, logger)
-	counter, err := c.CountActiveLoadTests(ctx)
+	counter, _, err := c.CountExistingLoadtests()
```
fabioserra, line 385: I think we are missing some test cases for this function :D 
- Tests for the new counter for load test types
- Tests for all phases
- Negative test when there is error
- Test when the loadtype or phase is unknown

fabioserra (changes requested):

fabioserra (approved):

fabioserra (approved):

PR #251, opened 2022-12-22
Update the locust Reporting example

eliflores:
To change the conditions used in the example, format imports, and remove an unused variable. 
Additionally: 
* Fixed some typos and linting issues.
* Mentioned that using test data in locust load tests with Kangal is not supported.

### DIFF
--- a/docs/locust/README.md
+++ b/docs/locust/README.md
@@ -10,7 +10,7 @@ Locust is one of the load generators implemented in Kangal. It uses the official
 
 Kangal requires a .py testfile describing the test.
 
-For more information, check [Locust official website](https://locust.io/).
+For more information, check the [Locust official site](https://locust.io/).
 
 ## How it works
 You can create Locust load tests using Kangal Proxy.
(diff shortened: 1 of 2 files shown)


docs/locust/README.md
```diff
@@ -72,7 +72,7 @@ $ curl -X POST http://${KANGAL_PROXY_ADDRESS}/load-test \
   -F type=Locust
 ```
 
-In this last example, the test will run infinitely and no `targetURL` were specified.
+In this last example, the test will run infinitely and no `targetURL` was specified.
```
s-radyuk, line 75:
```suggestion
In this last example, the test will run infinitely and no `targetURL` is specified in request, since it's set in the test code.
```

s-radyuk (commented):
Maybe we could also say in explicitly that unlike JMeter and K6, Locust doesn't support providing testdata as a separate file. I remember someone was asking about it in slack

docs/locust/README.md
eliflores: Updated in [112d17c](https://github.com/hellofresh/kangal/pull/251/commits/112d17c15adbb37c7fc1b4b7e810d2bd63c113fd).

s-radyuk (approved):

PR #253, opened 2022-12-27
Test data compression for all backends

### LINKED ISSUE #229: Backend agnostic testData behaviour
I have noticed that the testData compression introduced by #177 is not implemented for some load generators (maybe it is only supported by the JMeter integration code). Does it make sense to extract this behaviour to Kangal itself instead of delegating to specific load generators/backends? Maybe the main engine could deflate the CSV before calling `backend.Sync`, for example. Conversely, there would need to be arranged some way to include the init container that the JMeter backend uses to inflate it back.

Would the project be willing to accept a PR on that vein? This change could potentially include some other capabilities like the code that splits the testData CSV between workers, of course. Are there any caveats or potential roadblocks I should be aware of?

thiagoarrais:
This PR enables the test data compression available in JMeter for all other backends. This is achieved by compressing testdata prior to storing LoadTest objects _outside of_ the backend (i.e. should work for all future backends). It also externalizes the CSV splitting behaviour found in JMeter. To achieve that a new ` UsesCSVTestData`  property has been added to the backend interface.

Since the data is now gzipped, all testdata-using backends now need init containers to inflate back the data and make it available for the main container.

### Caveat 1: Generated mock

I couldn't figure out the right `mockgen`  incantation that would generate a conforming `mock.go`  and ended up changing it manually. Just point me in the right direction and I'll fix that.

### Caveat 2: CSV splitting

One may notice that I started assuming that k6 also used CSV test data and progressed to port CSV splitting for it. I later realised that JMeter is currently the only backend that support CSV test data exclusively. K6 and locust also support it, but just because they support any file type opaquely.

Maybe CSV splitting is just making the common code more complex. Just let me know if you think it really belongs inside the JMeter backend and I'll patch the PR accordingly.

Fixes #229

### DIFF
--- a/pkg/backends/backend.go
+++ b/pkg/backends/backend.go
@@ -17,11 +17,15 @@ type Backend interface {
 	// Type must return backend unique type string, called by both Proxy and Controller
 	Type() loadTestV1.LoadTestType
 
+	// UsesCSVTestData must signal whether the testdata files can be assumed to be CSV
+	// This is used to determine if testdata should be split between workers
+	UsesCSVTestData() bool
+
 	// TransformLoadTestSpec should validate and transform LoadTestSpec, called by Proxy
 	TransformLoadTestSpec(spec *loadTestV1.LoadTestSpec) error
 
 	// Sync should create resources if not exists, called by Controller
-	Sync(ctx context.Context, loadTest loadTestV1.LoadTest, reportURL string) error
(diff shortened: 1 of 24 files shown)


thiagoarrais:
By the way, shoutout to @amandahla, author of the [original gzip PR](https://github.com/hellofresh/kangal/pull/177) for the JMeter backend. Most of the code here was shamelessly copy-pasted from her.

PR #265, opened 2023-07-12
Fix/update opentelemetry instrumentation

st3veV:
Update opentelemetry instrumentation


s-radyuk (approved):
Great thanks for your contribution!

PR #272, opened 2023-09-06
Ensure all images can be modified with environment variables

lucasmdrs:
Use `envconfig` to set defaults and ensure all images used can be defined with environment variables


fabioserra (commented):
Please update the documentation for the env variables for locust
https://github.com/hellofresh/kangal/blob/master/docs/env-vars.md?plain=1#L57

fabioserra (approved):

PR #278, opened 2023-10-09
Don't allow different image tags for JMeter master and worker Docker images

s-radyuk:
**Current behaviour:** 
Kangal allows to use custom docker images when creating a loadtest object. For JMeter backend docker images for master and worker pods are defined separately. But they should still have the same tag to avoid inconsistency when running master and worker with different JMeter versions.

**Solution:**
This PR removes the usage of separate `MasterImageTag` and `WorkerImageTag` in code. Now the same tag will be used for both master and worker pods.

### DIFF
--- a/charts/kangal/Chart.yaml
+++ b/charts/kangal/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - performance tests
   - tests runner
 name: kangal
-version: 2.3.0
+version: 2.4.0
 home: https://github.com/hellofresh/kangal
 icon: https://raw.githubusercontent.com/hellofresh/kangal/master/logo.svg
 maintainers:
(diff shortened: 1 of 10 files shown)


fabioserra (approved):

lucasmdrs:
This change might cause some problems for some Kangal users. :warning: 

Restricting the worker and master images to have the same tag would force the way they tag their custom images (if any) and not necessarily fix the issue of having different JMeter versions running.

Aside from that, for JMeter backend master and worker images are also allowed to be passed from the Proxy request or even directly in the LoadTest spec, which makes this issue way more complex than what it seems.

I recommend we leave things as it is, since it gives the users safe defaults but  flexibility when they need at their own responsibility to ensure compatibility.

s-radyuk:
After discussion we realized we don't have enough evidence of different usage patterns and can't really make assumptions on what is the best approach for users. We agreed to leave the current logic in place.

